Issue #13, opened 2026-05-18
[FEAT][executors]: design zero-copy weight synchronization bridge


Flink-ddd:
Implement a high-performance weight synchronization mechanism in rl_engine/executors/bridge.py. 

Build a shared-memory or fast-buffer transfer layer to synchronize updated weights from DeepSpeed to vLLM with millisecond-level latency.

inaniloquentee:
This is the PR: https://github.com/Flink-ddd/RL-Kernel/pull/60

PR #60, opened 2026-06-02
feat: zero copy weight bridge

inaniloquentee:
## Summary
- Rebase issue #13 on top of the baseline that includes issue 55 native logp API parity.
- Replace the guarded legacy CUDA IPC placeholder with a real PyTorch CUDA IPC manifest bridge, lifecycle tracking, handle reuse, and validation.
- Add production-aligned smoke coverage for legacy CUDA IPC and vLLM IPC hot-weight update through vLLM's public IPC update path.

## Validation
- `pre-commit run --all-files`
- `py -3.13 -m mypy --ignore-missing-imports rl_engine/`
- `py -3.13 -m pytest rl_engine/tests/test_dispatch.py -v`
- `py -3.13 -m pytest tests\test_op_accuracy.py tests\test_weight_sync_bridge.py -q`
- `py -3.13 -m compileall rl_engine tests benchmarks`
- `py -3.13 -m mkdocs build --strict -f mkdocs.yaml`

## Runtime Smoke Results
- WSL2 legacy CUDA IPC smoke reached PyTorch CUDA IPC handle reconstruction in a spawned child process, then blocked with `torch.AcceleratorError: CUDA error: invalid resource handle` from `torch.storage._new_shared_cuda`.
- WSL2 vLLM IPC hot-weight update reached vLLM 0.18.1 `IPCWeightTransferEngine.receive_weights` through `LLM.update_weights(request)`, then blocked on the same PyTorch `_new_shared_cuda` invalid resource handle.
- A PyTorch-native `mp.Queue` CUDA tensor control experiment on the same WSL2 runtime failed at the same `_new_shared_cuda` invalid resource handle, so the remaining blocker is the current WSL2/driver/runtime legacy CUDA IPC path rather than the bridge/request implementation.

### DIFF
--- a/rl_engine/executors/bridge.py
+++ b/rl_engine/executors/bridge.py
@@ -1041,7 +1041,7 @@ class _CUDAVMMDriverBackend:
             "cuMemImportFromShareableHandle": [c.POINTER(c.c_uint64), c.c_void_p, c.c_int],
             "cuMemcpyHtoD_v2": [c.c_uint64, c.c_void_p, c.c_size_t],
             "cuMemcpyDtoD_v2": [c.c_uint64, c.c_uint64, c.c_size_t],
-            "cuCtxSynchronize": [],
+            "cuMemcpyDtoDAsync_v2": [c.c_uint64, c.c_uint64, c.c_size_t, c.c_void_p],
         }.items():
             getattr(self.lib, name).argtypes = args
 
@@ -1266,19 +1266,37 @@ class _CUDAVMMDriverBackend:
         if handle:
             self.lib.cuMemRelease(int(handle))
 
-    def copy_tensor_to_address(self, destination_address: int, tensor: torch.Tensor) -> None:
+    def copy_tensor_to_address(
+        self,
+        destination_address: int,
+        tensor: torch.Tensor,
+        *,
+        stream: Optional[int] = None,
+    ) -> None:
         if tensor.device.type != "cuda":
             raise WeightBridgeUnavailableError("CUDA VMM copy requires a CUDA source tensor")
         if not tensor.is_contiguous():
             tensor = tensor.contiguous()
         nbytes = int(tensor.numel() * tensor.element_size())
+        if stream is None:
+            self._check(
+                self.lib.cuMemcpyDtoD_v2(
+                    int(destination_address),
+                    int(tensor.data_ptr()),
+                    nbytes,
+                ),
+                "cuMemcpyDtoD",
+            )
+            return
         self._check(
-            self.lib.cuMemcpyDtoD_v2(int(destination_address), int(tensor.data_ptr()), nbytes),
-            "cuMemcpyDtoD",
+            self.lib.cuMemcpyDtoDAsync_v2(
+                int(destination_address),
+                int(tensor.data_ptr()),
+                nbytes,
+                self.ctypes.c_void_p(int(stream)),
+            ),
+            "cuMemcpyDtoDAsync",
         )
-
-    def synchronize(self) -> None:
-        self._check(self.lib.cuCtxSynchronize(), "cuCtxSynchronize")
 
 
 class WeightBridgeError(RuntimeError):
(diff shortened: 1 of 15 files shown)


Flink-ddd:
Excellent work! can you add related doc and explain this function and user guide. meanwile, cc @bitborne @EthanZero2Hero PTAL

inaniloquentee:
> Excellent work! can you add related doc and explain this function and user guide. meanwile, cc @bitborne @EthanZero2Hero PTAL

Thanks for the feedback! Added related docs.

rl_engine/executors/bridge.py
```diff
@@ -1,65 +1,2600 @@
...
+        if not tensor.is_contiguous():
+            tensor = tensor.contiguous()
+        nbytes = int(tensor.numel() * tensor.element_size())
+        self._check(
+            self.lib.cuMemcpyDtoD_v2(int(destination_address), int(tensor.data_ptr()), nbytes),
+            "cuMemcpyDtoD",
+        )
+
+    def synchronize(self) -> None:
+        self._check(self.lib.cuCtxSynchronize(), "cuCtxSynchronize")
```
Flink-ddd: Please replace the blocking cuCtxSynchronize() with fine-grained stream synchronization using torch.cuda.Event(interprocess=True) to prevent stalling the entire GPU context and introducing severe pipeline bubbles.

Example Implementation:

```
# Publisher side Training
sync_event = torch.cuda.Event(interprocess=True)
sync_event.record(torch.cuda.current_stream())
ipc_event_handle = sync_event.ipc_handle()

# (Include `ipc_event_handle` in the manifest's metadata)

# Consumer side Inference
remote_event = torch.cuda.Event.from_ipc_handle(
    device=self.device_index, 
    handle=manifest.metadata["ipc_event_handle"]
)
remote_event.wait(torch.cuda.current_stream())
```

rl_engine/executors/bridge.py
```diff
@@ -1,65 +1,2600 @@
...
+        try:
+            for name, tensor in model.state_dict().items():
+                if not isinstance(tensor, torch.Tensor):
+                    continue
+                if tensor.device.type != "cpu":
+                    raise WeightBridgeUnavailableError(
+                        "SharedMemoryTensorBridge currently supports CPU tensors only. "
+                        f"Tensor {name} is on {tensor.device}."
+                    )
+                snapshot = tensor.detach().clone(memory_format=torch.preserve_format)
```
Flink-ddd: Using .clone() here doubles the VRAM usage during the weight publishing phase. For a 70B-class model, this means an instant ~140GB spike, which will easily cause an OOM and defeats the core purpose of a "Zero-Copy" bridge.

To maintain true Zero-Copy semantics, we should strictly rely on `tensor.detach().contiguous()` and use the IPC Event synchronization (discussed above) to ensure that the read/write lifecycles are safe, completely avoiding the memory penalty of deep copying.

**Suggested fix:**
```
# Remove the expensive .clone()
snapshot = tensor.detach().contiguous()
```

rl_engine/executors/bridge.py
```diff
@@ -1,65 +1,2600 @@
...
+            for name, descriptor in manifest.tensors.items()
+            for entry in [tensor_metadata[name]]
+        }
+
+    def import_model_weights(self, ipc_handles: Mapping[str, Any]) -> Mapping[str, torch.Tensor]:
+        tensors: dict[str, torch.Tensor] = {}
         for name, info in ipc_handles.items():
-            cuda_float_tensor = cast(Any, torch.cuda).FloatTensor
-            storage = cuda_float_tensor._new_shared_cuda(info["handle"])
-            tensor = cuda_float_tensor(storage).view(info["shape"])
```
Flink-ddd, line 62: There is a guaranteed NameError in IPCWeightBridge.import_model_weights. 

The variable remote_weights is being assigned to here, but it is never initialized in this scope. Since you are collecting tensors into the tensors dictionary below, this line should be completely removed to prevent a runtime crash.

Flink-ddd (changes requested):

Flink-ddd:
LGTM, Let's merge this PR first, if has issue in the furture, then focus handle it.

PR #86, opened 2026-06-08
[BUG] Fix temperature applied twice in FlashInfer sampling path

### LINKED ISSUE #81: [BUG] Temperature applied twice in FlashInfer sampling path
Thanks author for interesting project! But I found some bug(?) in SamplerBackend.

SamplerBackend.sample() divides logits by temperature once for all backends, then the FlashInfer branch divides by temperature a second time.

## Location
[rl_engine/kernels/sampling.py:42-68](https://github.com/RL-Align/RL-Kernel/blob/main/rl_engine/kernels/sampling.py#L42-L68)

```
def sample(self, logits, top_k=None, top_p=None, temperature=1.0, deterministic=True):
    logits = logits.contiguous()
    if temperature != 1.0:
        logits = logits / temperature          # division #1 (all backends)

    if self.backend == constants.BackendLib.FLASHINFER.value:
        ...
        logits = logits.float().contiguous()
        if temperature != 1.0:
            logits.div_(temperature)            # division #2 (FlashInfer only) ← BUG
        probs = torch.softmax(logits, dim=-1)
```

I can make a PR to fix it.

KJLdefeated:
SamplerBackend.sample() divided logits by temperature once for all backends, then divided again inside the FlashInfer branch, producing softmax(logits / temperature**2) for any temperature != 1.0.

Remove the redundant second division so temperature is applied exactly once. The PyTorch fallback path was already correct, so this only affects FlashInfer deployments; temperature == 1.0 was unaffected (both divisions were guarded), which is why default-temperature tests never caught it.

Add tests/test_sampler_temperature.py covering the FlashInfer path with a stubbed flashinfer module: the recorded probs must equal softmax(logits / temperature), and temperature == 1.0 must be unscaled.

fixes #81 

## Summary by CodeRabbit

* **Bug Fixes**
  * Ensure temperature scaling is applied exactly once during sampling so probability computations are correct.

* **Tests**
  * Added CPU regression tests that verify probabilities are computed correctly for scaled (temperature=2.0) and unscaled (temperature=1.0) scenarios.

### DIFF
--- a/rl_engine/kernels/sampling.py
+++ b/rl_engine/kernels/sampling.py
@@ -50,9 +50,8 @@ class SamplerBackend(nn.Module):
         if self.backend == constants.BackendLib.FLASHINFER.value:
             from flashinfer.sampling import top_k_renorm_probs, top_p_sampling_from_probs
 
+            # Temperature is already applied above; do not re-divide here.
             logits = logits.float().contiguous()
-            if temperature != 1.0:
-                logits.div_(temperature)
 
             probs = torch.softmax(logits, dim=-1)
 
--- a/rl_engine/kernels/sampling.py
+++ b/rl_engine/kernels/sampling.py
@@ -50,9 +50,6 @@ class SamplerBackend(nn.Module):
         if self.backend == constants.BackendLib.FLASHINFER.value:
             from flashinfer.sampling import top_k_renorm_probs, top_p_sampling_from_probs
 
-            # Temperature is already applied above; do not re-divide here.
-            logits = logits.float().contiguous()
-
             probs = torch.softmax(logits, dim=-1)
 
             if top_k is None and top_p is None:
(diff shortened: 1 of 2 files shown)


Flink-ddd:
cc @inaniloquentee @bitborne PTAL

Flink-ddd (changes requested):
Excellent catch on the double division bug. However, please restore the line: logits = logits.float().contiguous().

Although the extra .div_(temperature) is gone, we must keep the cast to FP32 before softmax to avoid numerical overflow (real-world logits are FP16/BF16), and ensure contiguous memory for the backend to avoid implicit copy overhead.

Please update your test logits to dtype=torch.float16 to cover the necessary casting logic.

KJLdefeated:
Thanks for remind, I have put it back

Flink-ddd:
Thank you for update, but please resolve CI error first. also cc @inaniloquentee PTAL

Flink-ddd (approved):
LGTM now, Thanks.

inaniloquentee (approved):

inaniloquentee:
Merged

Issue #46, opened 2026-05-31
[FEAT][kernels]: implement fused GRPO loss with in-place group reward normalization


Flink-ddd:
### Context:
For GRPO, computing the advantage requires calculating the mean and standard deviation of rewards within a specific generation group (responses corresponding to the same prompt) and then computing the policy gradient loss. Doing this in standard PyTorch involves excessive broadcasting, expanding, and memory allocation.

## Tasks:
1. Implement a fused CUDA/Triton kernel (grpo_loss.py) that accepts flat sequences, group boundary indices, and raw rewards.
2. Perform group-wise reward normalization mean/std entirely within Shared Memory or fast registers without materializing intermediate broadcasted tensors.
3. Compute the clipped surrogate objective and the KL penalty in the same kernel pass.
4. Return the final scalar loss and the backward gradient for the policy logits.

KJLdefeated:
I am interested in this issue and have idea about how to do it. Can I try to resolve this?

Flink-ddd:
sure, Thank you.

PR #93, opened 2026-06-09
[FEAT][kernels]: implement fused GRPO loss with in-place group reward normalization

KJLdefeated:
#46 

## Summary
Implements the GRPO loss as a dispatchable `grpo_loss` op, eliminating the broadcasting/allocation overhead of a naive PyTorch implementation.

1. **`NativeGRPOLossOp`**: group-wise reward normalization (mean / population std, `clamp_min`), clipped surrogate objective + k3 reference-KL over active tokens.
2. **`TritonGRPOLossOp`**: `_group_norm_kernel` (per-group reward stats in registers) + token-parallel `_grpo_fwd_kernel`/`_grpo_bwd_kernel` with an **analytic** backward w.r.t. per-token logps. Each token gathers its advantage on the fly (`seq_id = idx // completion_len`), so **no broadcasted `[B, T]` advantage tensor is ever materialized**. Wrapped in a `torch.autograd.Function`.
3. **Registry dispatch** — `grpo_loss` resolves to the Triton backend on CUDA/ROCm, PyTorch native on CPU / when Triton is unavailable.

## Tests (`tests/test_grpo_loss.py`)
Built on `rl_engine.testing` fixtures (`make_synthetic_rl_kernel_batch`, `selected_logprobs_reference`, reference helpers):
- native op vs the reference group-norm + surrogate/KL recipe;
- Triton forward / backward / grad-scaling / per-sequence-apply vs native;
- `logp → grpo_loss` pipeline (NativeLogpOp + dispatched CUDA fused logp), incl.
  differentiability to logits;
- loss-step: masked-token invariance + an SGD step that lowers the loss;
- registry dispatch.

## Benchmark (`benchmarks/benchmark_grpo_loss.py`)
| shape (P×S×L) | tokens | fwd speedup | fwd+bwd speedup | VRAM (native→triton) |
|---|---|---|---|---|
| 64×8×512   | 0.26M | 4.70× | 2.98× | 10 MB → 1 MB |
| 128×8×1024 | 1.05M | 3.16× | 2.66× | 40 MB → 4 MB |
| 256×16×1024| 4.19M | 2.97× | 2.99× | 160 MB → 16 MB |

~3–4.7× faster forward, ~3× forward+backward, ~10× less peak VRAM.

## Notes
- `old_logps`/`ref_logps` are precomputed per-token constants.
- Group normalization stays a separate lightweight kernel by design (scalar per sequence; fusing it into the token kernel would trade away token-parallelism).

## Summary by CodeRabbit

* **New Features**
  * Added GRPO (Group Relative Policy Optimization) loss with both high-performance CUDA (Triton) and native PyTorch backends; runtime selects the best available implementation.

* **Benchmarks**
  * Added a CLI benchmark tool measuring forward and forward+backward latency and peak VRAM across configurable shapes, outputting markdown tables.

* **Tests**
  * Added comprehensive tests for loss values, gradients, grouping modes, backend parity, masked tokens, and integration.

* **Documentation**
  * Added operator docs and index entry describing usage, tensor contracts, and performance notes.

### DIFF
--- a/rl_engine/kernels/ops/pytorch/loss/grpo_loss.py
+++ b/rl_engine/kernels/ops/pytorch/loss/grpo_loss.py
@@ -70,7 +70,7 @@ class NativeGRPOLossOp:
 
         means = sums / counts
         variance = (sq_sums / counts) - means * means
-        stds = variance.clamp_min(0.0).sqrt().clamp_min(eps)
+        stds = variance.clamp_min(eps**2).sqrt()
 
         return (flat_rewards - means[group_id]) / stds[group_id]
 
--- a/rl_engine/kernels/ops/pytorch/loss/grpo_loss.py
+++ b/rl_engine/kernels/ops/pytorch/loss/grpo_loss.py
@@ -1,0 +1,193 @@
+# SPDX-License-Identifier: Apache-2.0
+# Copyright (c) 2026 RL-Kernel Contributors
+
+from __future__ import annotations
+
+from typing import Optional, Sequence, Tuple
+
+import torch
+
+
+class NativeGRPOLossOp:
+    """Pure PyTorch native fallback for the fused GRPO loss."""
+
+    def __init__(self) -> None:
+        pass
+
+    def __call__(
+        self,
+        current_logps: torch.Tensor,
+        old_logps: torch.Tensor,
+        ref_logps: torch.Tensor,
+        rewards: torch.Tensor,
+        completion_mask: torch.Tensor,
+        *,
+        clip_eps: float = 0.2,
+        beta: float = 0.0,
+        samples_per_prompt: Optional[int] = None,
+        group_boundaries: Optional[Sequence[int] | torch.Tensor] = None,
+        eps: float = 1e-6,
+    ) -> Tuple[torch.Tensor, torch.Tensor, torch.Tensor]:
+        return self.forward(
+            current_logps,
+            old_logps,
+            ref_logps,
+            rewards,
+            completion_mask,
+            clip_eps=clip_eps,
+            beta=beta,
+            samples_per_prompt=samples_per_prompt,
+            group_boundaries=group_boundaries,
+            eps=eps,
+        )
+
+    def group_advantages(
+        self,
+        rewards: torch.Tensor,
+        *,
+        samples_per_prompt: Optional[int] = None,
+        group_boundaries: Optional[Sequence[int] | torch.Tensor] = None,
+        eps: float = 1e-6,
+    ) -> torch.Tensor:
+        """Normalize raw per-sequence rewards within each generation group."""
+        flat_rewards = rewards.reshape(-1).float()
+        num_sequences = flat_rewards.numel()
+        group_id = self._resolve_group_ids(
+            num_sequences,
+            device=flat_rewards.device,
+            samples_per_prompt=samples_per_prompt,
+            group_boundaries=group_boundaries,
+        )
+
+        num_groups = int(group_id.max().item()) + 1 if num_sequences else 0
+        counts = flat_rewards.new_zeros(num_groups).index_add_(
+            0, group_id, torch.ones_like(flat_rewards)
+        )
+        sums = flat_rewards.new_zeros(num_groups).index_add_(0, group_id, flat_rewards)
+        sq_sums = flat_rewards.new_zeros(num_groups).index_add_(
+            0, group_id, flat_rewards * flat_rewards
+        )
+
+        means = sums / counts
+        variance = (sq_sums / counts) - means * means
+        stds = variance.clamp_min(0.0).sqrt().clamp_min(eps)
+
+        return (flat_rewards - means[group_id]) / stds[group_id]
+
+    @staticmethod
+    def expand_advantages(
+        sample_advantages: torch.Tensor,
+        completion_mask: torch.Tensor,
+    ) -> torch.Tensor:
+        """Broadcast per-sequence advantages to per-token, zeroing masked tokens."""
+        bool_mask = completion_mask.bool()
+        expanded = sample_advantages.reshape(-1, 1).expand_as(bool_mask).clone()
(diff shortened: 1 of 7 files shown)


Flink-ddd:
Thank you for your contribution! Could you add a readme for this kernel to explain how to use it? You can create your operator in the docs/operators directory.

KJLdefeated:
@Flink-ddd Documents are added.

inaniloquentee:
Thanks for the contribution, this is a useful GRPO loss addition. One concern: the native path masks inactive tokens only after `exp`/surrogate/KL computation, so extreme padding values can create `inf`/`nan` in the autograd graph and leak `nan` gradients. Could we sanitize masked positions or operate only on active tokens before those numerically sensitive ops?

KJLdefeated:
Thanks for remind! I will sanitize the exponents before exp, masked_fill on delta and diff, so masked positions get ratio=1 / kl=0 (finite).

inaniloquentee:
> Thanks for remind! I will sanitize the exponents before exp, masked_fill on delta and diff, so masked positions get ratio=1 / kl=0 (finite).

LGTM, Thank you for your contribution!

inaniloquentee (approved):

rl_engine/kernels/ops/pytorch/loss/grpo_loss.py
```diff
@@ -0,0 +1,194 @@
...
+            0, group_id, torch.ones_like(flat_rewards)
+        )
+        sums = flat_rewards.new_zeros(num_groups).index_add_(0, group_id, flat_rewards)
+        sq_sums = flat_rewards.new_zeros(num_groups).index_add_(
+            0, group_id, flat_rewards * flat_rewards
+        )
+
+        means = sums / counts
+        variance = (sq_sums / counts) - means * means
+        stds = variance.clamp_min(0.0).sqrt().clamp_min(eps)
```
Flink-ddd: Great implementation, but there's a classic PyTorch trap here:

stds = variance.clamp_min(0.0).sqrt().clamp_min(eps)

While rewards are currently detached, if we ever pass differentiable rewards from a Critic model in the future, backpropagating through sqrt(0) yields NaN gradients. It is safer to clamp the variance before the square root to a small epsilon.

Suggestion:

`stds = variance.clamp_min(eps**2).sqrt()`

This guarantees strictly positive inputs to sqrt both in forward and backward passes.

rl_engine/kernels/ops/triton/triton_grpo_loss.py
```diff
@@ -0,0 +1,359 @@
...
+    return 1 if x <= 1 else 1 << (x - 1).bit_length()
+
+
+@triton.jit
+def _group_norm_kernel(
+    rewards_ptr,
+    bounds_ptr,  # int32[num_groups + 1], CSR-style group offsets
+    adv_ptr,  # float32[N], per-sequence advantages (output)
+    eps,
+    GROUP_BLOCK: tl.constexpr,
```
Flink-ddd, line 36: in _group_norm_kernel launch config

I noticed you are launching _group_norm_kernel with GROUP_BLOCK = _next_pow2(max_group).

While this works perfectly for standard PPO/GRPO setups where samples_per_prompt is small (e.g., 8, 16, or 64), it will crash or fail to compile if a user provides highly skewed group_boundaries resulting in a max_group > 2048. Triton has hardware limits on block sizes.

Suggestion:
For this PR, we can keep the current design but please add an explicit check and throw a ValueError in TritonGRPOLossOp._build_bounds if max_group > 1024.

For future scaling, we might need a tiled reduction kernel here, but a safety assertion is enough for now to prevent cryptic CUDA errors.

rl_engine/kernels/ops/triton/triton_grpo_loss.py
```diff
@@ -0,0 +1,359 @@
...
+    clipped = tl.minimum(tl.maximum(ratio, lo), hi) * adv
+    policy_term = -tl.minimum(unclipped, clipped)
+
+    diff = ref - cur
+    kl_term = tl.exp(diff) - diff - 1.0
+
+    policy_term = tl.where(keep, policy_term, 0.0)
+    kl_term = tl.where(keep, kl_term, 0.0)
+
+    tl.atomic_add(partials_ptr + 0, tl.sum(policy_term, axis=0))
```
Flink-ddd: Atomic Contention in _grpo_fwd_kernel

tl.atomic_add(partials_ptr + 0, tl.sum(policy_term, axis=0))
tl.atomic_add(partials_ptr + 1, tl.sum(kl_term, axis=0))

Having all SM blocks hammer the exact same 2 global memory addresses with atomic adds will cause severe memory contention on high-end GPUs like H100s.

Suggestion:
Instead of using tl.atomic_add, it is generally much faster to have the kernel output a [grid_size] tensor of block-level sums, and then perform a simple torch.sum() over it in PyTorch (in _GRPOLossFunction.forward).

Since this only happens once per forward pass, it’s not a blocker for merging, but consider changing this if you plan to submit further optimizations! I'm happy to approve this PR once the sqrt numerical trap in the native op is addressed.

Flink-ddd (changes requested):
Awesome work and great memory optimizations! Just a few numerical and performance edge cases to address before merging.

rl_engine/kernels/ops/pytorch/loss/grpo_loss.py
KJLdefeated: I see. Thanks for the suggestion! I did not notice that.

rl_engine/kernels/ops/triton/triton_grpo_loss.py
KJLdefeated: Ok, I will use try-except to catch it

KJLdefeated:
@Flink-ddd Thx for suggestion! The requested changes are done.

Flink-ddd (approved):
LGTM now, Thank you for your contribution.

Issue #47, opened 2026-05-31
[FEAT][executors]: implement stateless (Zero-KV-Cache) forward engine for Reference and Reward models


Flink-ddd:
### Context:
In RLHF pipelines, Reference Models and Reward Models do not generate tokens; they only perform a single forward pass to compute log-probabilities or scalar scores. Instantiating them with standard generation engines like full vLLM pre-allocates massive amounts of VRAM for the Paged KV Cache, which goes completely unused.

## Tasks:
1. Create a lightweight execution wrapper stateless_executor.py specifically for Reference/Reward scoring.
2. Bypass the standard Paged Attention / KV Cache allocation mechanisms.
3. Force the engine to use standard FlashAttention-2 (without caching) for these models, maximizing the VRAM available for the Policy model's batch size.

inaniloquentee:
This is the PR: https://github.com/RL-Align/RL-Kernel/pull/94

PR #94, opened 2026-06-10
[FEAT][executors]: implement stateless zero-KV scoring executor

inaniloquentee:
## Summary
- Add `StatelessForwardExecutor` for no-cache reference/reward scoring.
- Add a local paged-KV scoring baseline and smoke benchmark for footprint comparison.
- Wire stateless rewards/reference logprobs into rollout payloads and trainer batch construction.
- Add unit and HF integration coverage for the new scoring path.

## Validation
- DCO: commit includes `Signed-off-by: inaniloquentee <3051000145@qq.com>`
- `py -3.13 -m pre_commit run --all-files`
- `py -3.13 -m mypy --ignore-missing-imports rl_engine/`
- `py -3.13 -m pytest rl_engine/tests/test_dispatch.py -v`
- `py -3.13 -m mkdocs build --strict -f mkdocs.yaml`
- `py -3.13 -m pytest -q` (`174 passed, 1 skipped`)
- `py -3.13 benchmarks\benchmark_stateless_executor.py --smoke --mode both --compare-paged-kv`

## Notes
- Real vLLM runtime is not available in the local `cp313/win_amd64` environment, so the benchmark includes a reproducible local paged-KV comparison path.

## Summary by CodeRabbit

* **New Features**
  * Added a stateless scoring executor supporting reference log-probabilities, rewards, and combined scoring.
  * Added a paged KV reservation scoring baseline with reservation/cache metrics.
  * Added a CLI benchmark tool for stateless scoring with optional paged-KV comparison.
  * Integrated stateless scoring into the training pipeline with grouped candidates and payload-backed reward/log-prob handling.
* **Bug Fixes**
  * Improved KL reference log-prob computation and advantage metrics to reflect only active completion tokens.
* **Tests**
  * Added unit/integration coverage for stateless scoring, paged KV reservation, and heavy-runtime import isolation.

### DIFF
--- a/benchmarks/benchmark_stateless_executor.py
+++ b/benchmarks/benchmark_stateless_executor.py
@@ -1,0 +1,481 @@
+# SPDX-License-Identifier: Apache-2.0
+# Copyright (c) 2026 RL-Kernel Contributors
+
+from __future__ import annotations
+
+import argparse
+import csv
+import json
+import sys
+from datetime import datetime, timezone
+from pathlib import Path
+from types import SimpleNamespace
+from typing import Any
+
+import torch
+
+REPO_ROOT = Path(__file__).resolve().parents[1]
+if str(REPO_ROOT) not in sys.path:
+    sys.path.insert(0, str(REPO_ROOT))
+
+from rl_engine.executors.paged_kv_baseline import (  # noqa: E402
+    PagedKVScoringBaseline,
+    PagedKVScoringConfig,
+)
+from rl_engine.executors.stateless_executor import (  # noqa: E402
+    StatelessForwardConfig,
+    StatelessForwardExecutor,
+    StatelessForwardInputs,
+)
+
+CSV_COLUMNS = [
+    "timestamp",
+    "candidate",
+    "mode",
+    "stage",
+    "batch_size",
+    "seq_len",
+    "active_tokens",
+    "device",
+    "dtype",
+    "elapsed_ms",
+    "peak_allocated_mb",
+    "peak_reserved_mb",
+    "paged_kv_baseline_mb",
+    "kv_cache_output_mb",
+    "zero_kv_cache_savings_mb",
+    "status",
+    "notes",
+]
+
+SCORING_MODES = {"reference", "reward", "both"}
+DTYPE_CHOICES = {
+    "float16": torch.float16,
+    "bfloat16": torch.bfloat16,
+    "float32": torch.float32,
+}
+
+
+class TinyReferenceModel(torch.nn.Module):
+    """Small deterministic causal-LM-like module for executor smoke benchmarks."""
+
+    def __init__(self, vocab_size: int, hidden_dim: int):
+        super().__init__()
+        self.embedding = torch.nn.Embedding(vocab_size, hidden_dim)
+        self.proj = torch.nn.Linear(hidden_dim, vocab_size)
+        self.use_cache_calls: list[bool | None] = []
+
+    def forward(self, input_ids, attention_mask=None, use_cache=None):
+        del attention_mask
+        self.use_cache_calls.append(use_cache)
+        hidden = self.embedding(input_ids.long())
+        return SimpleNamespace(
+            logits=self.proj(hidden),
+            past_key_values=_tiny_past_key_values(hidden, use_cache),
+        )
+
+
+class TinyRewardModel(torch.nn.Module):
+    """Small deterministic scalar-reward module for executor smoke benchmarks."""
+
+    def __init__(self, vocab_size: int, hidden_dim: int):
+        super().__init__()
+        self.embedding = torch.nn.Embedding(vocab_size, hidden_dim)
+        self.reward_head = torch.nn.Linear(hidden_dim, 1)
+        self.use_cache_calls: list[bool | None] = []
+
+    def forward(self, input_ids, attention_mask=None, use_cache=None):
+        self.use_cache_calls.append(use_cache)
+        hidden = self.embedding(input_ids.long())
+        mask = attention_mask.to(device=hidden.device, dtype=hidden.dtype).unsqueeze(-1)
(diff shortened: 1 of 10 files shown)


Flink-ddd:
Excellent work! Thank you for your contribution, Could you add an issue to this PR? If this issue exists, please do so. otherwise, please create one.

inaniloquentee:
> Excellent work! Thank you for your contribution, Could you add an issue to this PR? If this issue exists, please do so. otherwise, please create one.

This PR already has a corresponding issue: https://github.com/RL-Align/RL-Kernel/issues/47

rl_engine/executors/stateless_executor.py
```diff
@@ -0,0 +1,564 @@
...
+        temperature=temperature,
+        output_dtype=output_dtype,
+    )
+    result = torch.zeros(
+        inputs.input_ids.shape,
+        device=logits.device,
+        dtype=output_dtype,
+    )
+    result[:, 1:] = shifted_logps
+    return result.masked_fill(~_bool_mask(inputs.completion_mask, device=result.device), 0.0)
```
Flink-ddd, line 253: In score_reference_logprobs, result[:, 0] is always 0 but the un-shifted completion_mask is applied afterward, so any sequence with completion_mask[:, 0] == True gets its logprob silently zeroed. This is an undocumented precondition. Suggest adding to _validate_inputs:

```
if bool(_bool_mask(inputs.completion_mask, device=input_ids.device)[:, 0].any().item()):
    raise ValueError("completion_mask[:, 0] must be False; completions cannot start at position 0")
```

rl_engine/executors/stateless_executor.py
```diff
@@ -0,0 +1,564 @@
...
+
+    use_cache_targets = 0
+    attention_targets = 0
+
+    for target in _model_config_targets(model):
+        if hasattr(target, "use_cache"):
+            target.use_cache = False
+            use_cache_targets += 1
+        if config.attention_backend != "model_default":
+            target.attn_implementation = config.attention_backend
```
Flink-ddd, line 355: configure_stateless_model mutates model.config (and generation_config) in place permanently — if the same model instance is reused elsewhere its attention backend / use_cache stay overwritten with no restore. Either document this side effect, or wrap score() in a restore-on-exit context manager:

```
from contextlib import contextmanager
_MISSING = object()

@contextmanager
def _temporarily_configure(model, config):
    saved = [
        (t, getattr(t, "use_cache", _MISSING), getattr(t, "_attn_implementation", _MISSING))
        for t in _model_config_targets(model)
    ]
    try:
        configure_stateless_model(model, config)
        yield
    finally:
        for t, uc, attn in saved:
            if uc is not _MISSING:
                t.use_cache = uc
            if attn is not _MISSING:
                t._attn_implementation = attn
```

rl_engine/executors/stateless_executor.py
```diff
@@ -0,0 +1,564 @@
...
+        return False
+    if isinstance(exc, (ImportError, ModuleNotFoundError, StopIteration)):
+        return True
+    message = str(exc).lower()
+    markers = (
+        "flash_attention",
+        "flash attention",
+        "flash-attn",
+        "flash_attn",
+        "kernels",
```
Flink-ddd: In _should_fallback_attention_backend, the marker "kernels" is too broad, it will match unrelated errors (e.g. a CUDA kernel launch failure) and silently fall back to eager, hiding the real exception. Remove "kernels" or tighten it to something like "flash attention kernels".

rl_engine/executors/deepspeed_trainer.py
```diff
@@ -145,6 +147,16 @@ def train(self, rollout: RolloutStageResult) -> TrainingStageResult:
                 "training_device": str(self.device),
                 "deepspeed_engine": type(self.engine).__name__,
                 "deepspeed_zero_stage": self.config.zero_stage,
+                "advantage_mean": (
+                    float(active_advantages.mean().detach().cpu().item())
+                    if active_advantages.numel()
+                    else 0.0
+                ),
+                "advantage_std": (
```
Flink-ddd: Advantages are already group-normalized in _compute_group_relative_advantages, so taking mean/std again over all flattened active tokens is neither the per-group std nor a standard GRPO statistic. unbiased=False is fine, but rename to make scope explicit, e.g. active_advantage_std_global, to avoid misreading.

benchmarks/benchmark_stateless_executor.py
```diff
@@ -0,0 +1,481 @@
...
+            f"zero_kv_cache={result.metrics['zero_kv_cache']};"
+            f"attention_backend={result.metrics['attention_backend']};"
+            f"hidden_dim={args.hidden_dim};vocab_size={args.vocab_size};{_environment()}"
+        ),
+    }
+    if args.compare_paged_kv:
+        paged_result = _score_paged_kv(args, inputs, model)
+        paged_metrics = paged_result.metrics
+        row["paged_kv_baseline_mb"] = f"{float(paged_metrics['paged_kv_cache_reserved_mb']):.4f}"
+        savings_mb = float(paged_metrics["total_kv_cache_mb"]) - float(
```
Flink-ddd, line 266: The zero_kv_cache_savings_mb value in _run measures savings vs. a synthetic local reservation, not real vLLM pre-allocation (the PR description acknowledges this). Suggest renaming the column in CSV_COLUMNS to savings_vs_local_reservation_baseline so it isn't reported as real vLLM savings.

Flink-ddd (changes requested):
Good approach and test coverage, happy to re-review once fixed.

inaniloquentee:
> Good approach and test coverage, happy to re-review once fixed.

Thanks for the careful review! I’ve addressed the requested changes and added/updated coverage for them. The relevant tests are passing now.

Flink-ddd (approved):
LGTM now, Thank you for your update.

bitborne (approved):
LGTM！

frank-2077:
LGTM！

frank-2077 (approved):

Issue #40, opened 2026-05-31
[FEAT][kernels]: implement Fused Policy Ratio and KL Penalty kernel (Triton/CUDA/ROCm)


Flink-ddd:
### Context:
In our RL alignment pipelines (PPO/GRPO), computing the policy ratio and the KL divergence penalty introduces significant memory bandwidth overhead due to the creation of intermediate tensors and active token masking. We need a fused operator to compute these metrics in a single kernel launch.

## Tasks:
1. Implement a fused Triton kernel in rl_engine/kernels/ops/triton/loss/ that takes policy_logits, ref_logits, action_ids, and attention_mask as inputs.
2. The kernel should return the policy_ratio and kl_penalty for active tokens directly, bypassing intermediate log-softmax materialization.
3. Integrate this operator into the upcoming GRPO and PPO loss function wrappers.

KJLdefeated:
Hi @Flink-ddd , I think I can resolve this.

Flink-ddd:
sure, Thanks.

KJLdefeated:
Hi @Flink-ddd , let me clarify something: the current logp kernel(csrc/fused_logp_kernel.cu) did not have backward function? If it is true, do we need one?

Flink-ddd:
this file create by @inaniloquentee and details see pr #33, if true, please add one. cc @inaniloquentee PTAL.

inaniloquentee:
> Hi [@Flink-ddd](https://github.com/Flink-ddd) , let me clarify something: the current logp kernel(csrc/fused_logp_kernel.cu) did not have backward function? If it is true, do we need one?

You're right, there's no backward function yet. @KJLdefeated feel free to add one.

KJLdefeated:
Ok I will first implement fused logp + policy ratio and KL kernel (fwd and bwd) first. For CUDA logp bwd kernel, I will launch another PR

PR #97, opened 2026-06-11
[FEAT][kernels]: implement Fused Policy Ratio and KL Penalty kernel

KJLdefeated:
#40 

## Summary
- Adds a new fused Policy Ratio + KL Penalty (ratio_kl) operator that turns raw logits into the per-token PPO/GRPO importance ratio and reference-KL penalty via online softmax. 
- Because grpo_loss use same ratio and kl computation kernel, I refactors grpo_loss (#93 ) to consume logits and build on this operator, deleting its bespoke forward/backward Triton kernels (group norm kernel is kept). Future PPO Loss can also use this kernel.
- The ratio_kl forward peak VRAM is independent of vocab size; the fused path is 5–10× faster and trades GBs of log-softmax scratch for ~0, with the win growing as vocab grows.

### New operator — `ratio_kl`
- **`rl_engine/kernels/ops/triton/loss/ratio_kl.py`**: 
    - New Triton kernel + `_RatioKLFunction` autograd + `TritonRatioKLOp`. 
    - Streams an online LSE over the vocab in `BLOCK_V` tiles for policy and ref
- **`rl_engine/kernels/ops/pytorch/loss/ratio_kl.py`**: `NativeRatioKLOp` reference, reusing `NativeLogpOp`
- **`tests/test_ratio_kl.py`**:  native-vs-reference, masked neutrality, ratio==1 when `old==policy`, grad-to-policy-only, Triton fwd/bwd vs native (V=64 and 50257), registry dispatch.
- **`benchmarks/benchmark_ratio_kl.py`** : New benchmark (latency + peak VRAM, vocab sweep). 
- **`docs/operators/ratio-kl.md`**: operator doc.

### GRPO loss refactor — compose `ratio_kl` instead of a bespoke kernel
- **`rl_engine/kernels/ops/triton/{triton_grpo_loss.py → loss/grpo_loss.py}`**: 
    - Moved into `loss/`; 
    - Dropped `_grpo_fwd/_grpo_bwd` kernels and `_GRPOLossFunction`; 
    - Now calls `TritonRatioKLOp` and keeps only `_group_norm_kernel` for reward normalization. 
    - Make cleaner code.
- **`benchmarks/benchmark_grpo_loss.py`**: Updated to logits API
- **`tests/test_grpo_loss.py`**: Loss/grad/masking/SGD tests reworked for logits API
- **`docs/operators/grpo-loss.md`**

### Wiring
- `rl_engine/kernels/registry.py`: Register `ratio_kl` (Triton/PyTorch) for cuda/rocm/cpu; repoint `TRITON_GRPO_LOSS` to `triton.loss.grpo_loss`

## Performance (fp16, B=32, T=256)

| shape (P×S×L×V) | fwd speedup | fwd+bwd speedup | peak fwd VRAM (native → Triton) |
| --- | --- | --- | --- |
| 4×8×256×32768 | 5.2× | 2.8× | 2048 MB → ~0 MB |
| 4×8×256×50257 | 7.3× | 2.4× | 3141 MB → ~0 MB |
| 4×8×256×131072 | 10.3× | 3.4× | 8192 MB → ~0 MB |

## Notes

- PPO Wrapper will be implemented in the future
- The current CUDA Logp kernel is lacking backward function.

## Release Notes

* **New Features**
  * Introduced a fused `ratio_kl` operator for computing policy ratio and KL penalty in PPO/GRPO workflows.
  * Updated GRPO Loss to use policy/reference logits plus action IDs, deriving ratio and KL via the fused `ratio_kl` path.

* **Documentation**
  * Updated GRPO Loss operator docs to reflect the new logits-based API and gradient behavior.
  * Added new `ratio_kl` operator documentation and linked it from the operators index.

* **Tests**
  * Added a dedicated `ratio_kl` test suite (correctness, masking neutrality, gradient flow, and backend dispatch).
  * Refreshed GRPO Loss tests to use logits-based inputs and updated gradient expectations.

* **Benchmarking**
  * Updated benchmarks to reflect the new `ratio_kl`-driven composition and updated tensor shapes/CLI options.

### DIFF
--- a/rl_engine/kernels/ops/pytorch/loss/ratio_kl.py
+++ b/rl_engine/kernels/ops/pytorch/loss/ratio_kl.py
@@ -1,0 +1,51 @@
+# SPDX-License-Identifier: Apache-2.0
+# Copyright (c) 2026 RL-Kernel Contributors
+
+from __future__ import annotations
+
+from typing import Tuple
+
+import torch
+
+from rl_engine.kernels.ops.pytorch.loss.logp import NativeLogpOp
+
+
+class NativeRatioKLOp:
+    """PyTorch native fallback for the fused ratio + KL operator."""
+
+    def __init__(self) -> None:
+        self._logp = NativeLogpOp()
+
+    def __call__(
+        self,
+        policy_logits: torch.Tensor,
+        ref_logits: torch.Tensor,
+        action_ids: torch.Tensor,
+        attention_mask: torch.Tensor,
+        old_logps: torch.Tensor,
+    ) -> Tuple[torch.Tensor, torch.Tensor]:
+        return self.forward(policy_logits, ref_logits, action_ids, attention_mask, old_logps)
+
+    def _selected_logp(self, logits: torch.Tensor, action_ids: torch.Tensor) -> torch.Tensor:
+        # Clamp ids first so masked/pad positions never index out of bounds
+        safe_ids = action_ids.clamp(0, logits.size(-1) - 1).long()
+        return self._logp.apply_fp32(logits, safe_ids)
+
+    def forward(
+        self,
+        policy_logits: torch.Tensor,
+        ref_logits: torch.Tensor,
+        action_ids: torch.Tensor,
+        attention_mask: torch.Tensor,
+        old_logps: torch.Tensor,
+    ) -> Tuple[torch.Tensor, torch.Tensor]:
+        mask = attention_mask.to(torch.bool)
+        logp_policy = self._selected_logp(policy_logits, action_ids)
+        with torch.no_grad():
+            logp_ref = self._selected_logp(ref_logits, action_ids)
+
+        delta = (logp_policy - old_logps.float()).masked_fill(~mask, 0.0)
+        diff = (logp_ref - logp_policy).masked_fill(~mask, 0.0)
+        ratio = torch.exp(delta)
+        kl = torch.exp(diff) - diff - 1.0
+        return ratio, kl
--- a/rl_engine/kernels/ops/pytorch/loss/ratio_kl.py
+++ b/rl_engine/kernels/ops/pytorch/loss/ratio_kl.py
@@ -27,9 +27,7 @@ class NativeRatioKLOp:
         return self.forward(policy_logits, ref_logits, action_ids, attention_mask, old_logps)
 
     def _selected_logp(self, logits: torch.Tensor, action_ids: torch.Tensor) -> torch.Tensor:
-        # Clamp ids first so masked/pad positions never index out of bounds
-        safe_ids = action_ids.clamp(0, logits.size(-1) - 1).long()
-        return self._logp.apply_fp32(logits, safe_ids)
+        return self._logp.apply_fp32(logits, action_ids)
 
     def forward(
         self,
@@ -40,6 +38,13 @@ class NativeRatioKLOp:
         old_logps: torch.Tensor,
     ) -> Tuple[torch.Tensor, torch.Tensor]:
         mask = attention_mask.to(torch.bool)
+        vocab_size = policy_logits.size(-1)
+        invalid = mask & ((action_ids < 0) | (action_ids >= vocab_size))
+        if invalid.any():
+            raise ValueError(
+                f"action_ids at active (unmasked) positions must be in "
+                f"[0, {vocab_size}); found out-of-range ids."
+            )
         logp_policy = self._selected_logp(policy_logits, action_ids)
         with torch.no_grad():
             logp_ref = self._selected_logp(ref_logits, action_ids)
--- a/rl_engine/kernels/ops/triton/loss/ratio_kl.py
+++ b/rl_engine/kernels/ops/triton/loss/ratio_kl.py
@@ -1,0 +1,185 @@
+# SPDX-License-Identifier: Apache-2.0
+# Copyright (c) 2026 RL-Kernel Contributors
+
+"""Fused Policy-ratio + KL-penalty Triton operator (PPO/GRPO front-end).
+policy_ratio = exp(logp_policy - old_logp)
+kl_penalty   = exp(d) - d - 1,   d = logp_ref - logp_policy   (k3 estimator)
+"""
+
+from __future__ import annotations
+
+from typing import Tuple
+
+import torch
+import triton
+import triton.language as tl
+
+# Tile width
+_MAX_BLOCK_V = 2048
+
+
+@triton.jit
+def _ratio_kl_fwd_kernel(
+    policy_ptr,
+    ref_ptr,
+    action_ptr,
+    mask_ptr,
+    old_ptr,
+    ratio_ptr,
+    kl_ptr,
+    diff_ptr,  # saved for backward: d = logp_ref - logp_policy
+    logz_ptr,  # saved for backward: log-sum-exp of policy logits
+    V,
+    BLOCK_V: tl.constexpr,
+):
+    row = tl.program_id(0)
+    active = tl.load(mask_ptr + row) != 0
+
+    if active:
+        row_off = row.to(tl.int64) * V
+        a = tl.load(action_ptr + row)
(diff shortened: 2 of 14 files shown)


rl_engine/kernels/ops/triton/loss/ratio_kl.py
```diff
@@ -0,0 +1,185 @@
...
+class _RatioKLFunction(torch.autograd.Function):
+    @staticmethod
+    def forward(ctx, policy_logits, ref_logits, action_ids, attention_mask, old_logps):
+        if not policy_logits.is_cuda:
+            raise RuntimeError("TritonRatioKLOp requires CUDA/ROCm tensors.")
+        V = policy_logits.shape[-1]
+        pol = policy_logits.contiguous().view(-1, V)
+        ref = ref_logits.contiguous().view(-1, V)
+        n_rows = pol.shape[0]
+        act = action_ids.contiguous().view(-1).to(torch.int64)
```
Flink-ddd: [blocking] action_ids not clamped before indexing, diverges from native op

The native op clamps action ids before gathering:

`safe_ids = action_ids.clamp(0, logits.size(-1) - 1).long()`

but the Triton path indexes policy_ptr + row_off + a directly. For an active row with an out-of-range action_id (corrupt/unexpected data), this is an illegal memory access rather than a clamped read — the two backends behave differently on the same input. Suggest clamping host-side to keep parity:

`act = action_ids.contiguous().view(-1).clamp_(0, V - 1).to(torch.int64)`

tests/test_ratio_kl.py
```diff
@@ -0,0 +1,194 @@
...
+    (r_t * gr + k_t * gk).sum().backward()
+
+    pol_n = policy_logits.clone().requires_grad_(True)
+    r_n, k_n = native(pol_n, ref_logits, action_ids, mask, old)
+    (r_n * gr + k_n * gk).sum().backward()
+
+    assert pol_t.grad is not None
+    assert torch.isfinite(pol_t.grad).all()
+    assert torch.allclose(pol_t.grad, pol_n.grad, atol=1e-4, rtol=1e-4)
+
```
Flink-ddd, line 147: [test gap] No assertion that ref_logits receives no gradient on the Triton path
test_native_gradient_flows_to_policy_logits asserts ref_logits.grad is None, but the Triton backward tests never set requires_grad_ on ref_logits, so the freeze isn't actually verified for the fused op. Suggest a symmetric test:

```
@requires_triton_cuda
def test_triton_no_grad_to_ref():
    fused = TritonRatioKLOp()
    pol, ref, act, mask, old = _inputs(seed=8, device="cuda")
    pol = pol.clone().requires_grad_(True)
    ref = ref.clone().requires_grad_(True)
    r, k = fused(pol, ref, act, mask, old)
    (r.sum() + k.sum()).backward()
    assert ref.grad is None
```

tests/test_ratio_kl.py
```diff
@@ -0,0 +1,194 @@
...
+    pert = policy_logits.clone()
+    pert[inactive] = 1000.0
+    pert_r, pert_k = fused(pert, ref_logits, action_ids, mask, old)
+
+    active = mask.to(torch.bool)
+    assert torch.allclose(base_r[active], pert_r[active], atol=1e-5)
+    assert torch.allclose(base_k[active], pert_k[active], atol=1e-5)
+    assert torch.allclose(pert_r[inactive], torch.ones_like(pert_r[inactive]))
+
+
```
Flink-ddd, line 198: [test gap] No coverage for out-of-range action_ids / clamp behavior
Related to the clamp divergence above. Once the behavior is settled (clamp on both paths), add a test that puts an out-of-range id at a masked position and asserts both backends stay finite and agree on active outputs:

```
@requires_triton_cuda
def test_triton_handles_oob_action_ids():
    native = NativeRatioKLOp()
    fused = TritonRatioKLOp()
    pol, ref, act, mask, old = _inputs(seed=9, device="cuda", valid_density=0.7)
    act = act.clone()
    inactive = ~mask.to(torch.bool)
    act[inactive] = _VOCAB + 999  # garbage id at masked positions
    r_t, k_t = fused(pol, ref, act, mask, old)
    r_n, k_n = native(pol, ref, act, mask, old)
    active = mask.to(torch.bool)
    assert torch.isfinite(r_t).all() and torch.isfinite(k_t).all()
    assert torch.allclose(r_t[active], r_n[active], atol=1e-4, rtol=1e-4)
    assert torch.allclose(k_t[active], k_n[active], atol=1e-4, rtol=1e-4)
```

Flink-ddd (changes requested):
Overall clean, online-softmax fwd and the backward chain rule look correct, nice to see the bespoke GRPO kernels removed. One blocking item (action_ids clamp parity with the native op) plus two test-coverage suggestions inline. Happy to approve once the clamp is addressed.

KJLdefeated:
@Flink-ddd The requests have been resolved. Thx for reviews.

Flink-ddd (approved):
LGTM now, but also need @inaniloquentee to review.

Flink-ddd:
The GPU CI test is unrelated to the code because GPU resources are temporarily occupied, we'll try again later.

rl_engine/kernels/ops/triton/loss/ratio_kl.py
```diff
@@ -0,0 +1,185 @@
...
+            p = tl.load(policy_ptr + row_off + cols, mask=cmask, other=0.0).to(tl.float32)
+            soft = tl.exp(p - logz)
+            onehot = tl.where(cols == a, 1.0, 0.0)
+            grad = c * (onehot - soft)
+            tl.store(grad_policy_ptr + row_off + cols, grad, mask=cmask)
+
+
+class _RatioKLFunction(torch.autograd.Function):
+    @staticmethod
+    def forward(ctx, policy_logits, ref_logits, action_ids, attention_mask, old_logps):
```
inaniloquentee, line 127: Could we add a quick shape check before flattening here? If action_ids, attention_mask, or old_logps don't match policy_logits.shape[:-1], the Triton kernel can still launch and read the wrong rows.

rl_engine/kernels/ops/pytorch/loss/ratio_kl.py
```diff
@@ -0,0 +1,51 @@
...
+        ref_logits: torch.Tensor,
+        action_ids: torch.Tensor,
+        attention_mask: torch.Tensor,
+        old_logps: torch.Tensor,
+    ) -> Tuple[torch.Tensor, torch.Tensor]:
+        return self.forward(policy_logits, ref_logits, action_ids, attention_mask, old_logps)
+
+    def _selected_logp(self, logits: torch.Tensor, action_ids: torch.Tensor) -> torch.Tensor:
+        # Clamp ids first so masked/pad positions never index out of bounds
+        safe_ids = action_ids.clamp(0, logits.size(-1) - 1).long()
```
inaniloquentee: Non-blocking, but could we avoid clamping active token ids here? Clamping is fine for masked positions, but an out-of-range active action_id may be better surfaced as an error instead of silently using the last vocab entry.

inaniloquentee (changes requested):

rl_engine/kernels/ops/triton/loss/ratio_kl.py
KJLdefeated: Thx for suggestion, I will fix the requests.

rl_engine/kernels/ops/pytorch/loss/ratio_kl.py
KJLdefeated: I see, I will use a try-catch to deal with it

KJLdefeated:
@inaniloquentee the requests are resolved, thx for reviews!

inaniloquentee:
> @inaniloquentee the requests are resolved, thx for reviews!

LGTM now. Thanks!

inaniloquentee (approved):

Issue #96, opened 2026-06-11
Implement Batch-Invariant Deterministic LogProb CUDA Kernel to Eliminate Batch-Size Drift


inaniloquentee:
Implement a CUDA `Batch-Invariant Deterministic LogProb` kernel for token-level logprob computation. The same sequence must produce bitwise-identical logprobs regardless of batch size, batch position, padding layout, or other samples in the batch.

The goal is to eliminate train/inference mismatch caused by nondeterministic reduction order, preventing KL spikes and training collapse.

**Requirements**

- Do not use `atomicAdd` or any nondeterministic accumulation path.
- Use a fixed CUDA reduction topology:
  - fixed block/thread partitioning,
  - fixed tree-reduction order,
  - fixed reduction order for max, sum-exp, and final logprob computation.
- The output for the same sequence must be bitwise identical across:
  - different batch sizes,
  - different positions within the batch,
  - different padding/mask layouts,
  - mixed batches containing unrelated sequences.
- Accumulate in FP32 where needed. Inputs may support FP16 / BF16 / FP32.
- Do not rely on cuDNN/cuBLAS softmax implementations that may change behavior based on workspace, heuristics, or scheduling.
- Preserve existing logprob semantics, including masking, padding, and ignore-index behavior.

**Acceptance Criteria**

- Add a deterministic CUDA logprob kernel.
- Add batch-invariance tests:
  - place the same sequence in batches of different sizes,
  - move it to different batch positions,
  - mix it with unrelated sequences,
  - verify per-token and sequence logprobs are bitwise identical.
- Add repeatability tests showing identical output across repeated runs.
- Verify the kernel does not use `atomicAdd`.
- Compare numerics against the existing logprob path within an acceptable tolerance, while prioritizing deterministic stability.
- Leave the test structure ready for future ROCm and Triton parity checks.

inaniloquentee:
This is the PR: https://github.com/RL-Align/RL-Kernel/pull/98

PR #98, opened 2026-06-11
Add batch-invariant deterministic CUDA logp

inaniloquentee:
﻿## Summary

This PR adds an opt-in **batch-invariant deterministic CUDA selected-token logprob** backend for RL-Kernel. It is intended for RL train/eval paths where the same flattened row must produce the same selected logprob regardless of batch size, batch position, sparse indexing pattern, or unrelated rows in the launch.

## Why

GRPO/PPO-style training is sensitive to tiny logprob drift: policy ratios and KL penalties can amplify rollout-time vs train-time differences. Existing high-throughput reductions may change floating-point accumulation order with launch shape or sparse indexing. This PR adds a CUDA path whose reduction topology is locked by vocab bucket, so a row's result is independent of the surrounding batch while avoiding the worst cost of one universal block size.

## Implementation

- Adds `csrc/deterministic_logp_kernel.cu`.
- Uses one CUDA block per flattened logit row.
- Locks deterministic reduction topology by `vocab_size` only:
  - `vocab_size <= 128` uses `BlockSize=128`
  - `129 <= vocab_size <= 4096` uses `BlockSize=256`
  - `vocab_size > 4096` uses `BlockSize=512`
- Keeps bucket selection independent of batch size, sequence length, launch row count, sparse index count, and row position.
- Uses explicit two-pass log-sum-exp:
  - fixed tree reduction for row max
  - fixed tree reduction for exp-sum
- Avoids `atomicAdd`, CUB block reductions, autotuned block sizes, and dynamic reduction topology.
- Exposes dense, `out`, fp32, indexed `out`, and indexed fp32 bindings through `_C`.
- Adds `DeterministicLogpCUDAOp` plus registry op types:
  - `logp_deterministic`
  - `logp_deterministic_indexed`
- Adds user-facing aliases via `resolve_logp_op_type`, including `deterministic`, `deterministic_cuda`, and `batch_invariant`.
- Wires rollout/training config validation and `examples/grpo_single_gpu.py` through:
  - `--logp-backend`
  - `--require-batch-invariant-logp`
- Adds benchmark candidates:
  - `deterministic_fp32`
  - `deterministic_indexed_fp32`
- Removes default CUDA `--use_fast_math`; it is now opt-in with `KERNEL_ALIGN_USE_FAST_MATH=1`.
- Adds Windows CUDA build escape hatch for unsupported MSVC combinations via `KERNEL_ALIGN_ALLOW_UNSUPPORTED_MSVC=1`.

### CUDA Shape / Dtype / Bucket Matrix

| Axis | Values Covered | Assertion | Result |
| --- | --- | --- | --- |
| Vocab size | `1, 2, 31, 32, 33, 127, 128, 129, 255, 256, 257, 1024, 4095, 4096, 4097, 4099, 8192` | matches PyTorch `log_softmax(...).gather(...)` within dtype tolerance | Pass |
| Bucket boundaries | `128, 129, 4096, 4097` | dense and full-row indexed results are byte-for-byte invariant across batch sizes and positions | Pass |
| Block topology contract | source checks for `128/256/512` block constants and `128/4096` vocab limits | topology is fixed by vocab bucket and not autotuned | Pass |
| Batch/sequence shapes | `(1,1), (1,3), (2,5), (2,7), (3,4), (4,3), (4,5), (2,6), (2,3)` paired with vocab matrix | output shape equals token id shape | Pass |
| Input dtype | `fp16`, `bf16` when supported, `fp32` | fp32 deterministic output matches reference | Pass |
| `out` dtype | `fp16`, `bf16` when supported, `fp32`, `fp64` | reuses caller storage and matches reference after cast | Pass |
| Non-contiguous inputs | strided logits and strided token ids | wrapper handles non-contiguous tensors via reshape/contiguous prep | Pass |

### Determinism Matrix

| Case | Coverage | Assertion | Result |
| --- | --- | --- | --- |
| Repeatability | 20 repeated launches on same tensor | byte-for-byte identical output | Pass |
| Batch-size invariance | same target row packed into batch sizes `1, 2, 4, 8, 16` | target row byte-for-byte identical | Pass |
| Batch-position invariance | same target row placed at positions `0..7` | target row byte-for-byte identical | Pass |
| Batch-noise invariance | target row embedded in batch size `32`, unrelated rows randomized for seeds `20..29` | target row byte-for-byte identical | Pass |
| Dense/indexed parity | sparse row sets including single row, subset, duplicate rows, and all rows | indexed selected rows byte-match dense rows | Pass |

### Indexed / Edge-Case Matrix

| Case | Coverage | Expected Behavior | Result |
| --- | --- | --- | --- |
| Empty indices | `indexed_out` with zero row indices | output remains unchanged | Pass |
| Empty indices helper | `indexed_fp32` with zero row indices | returns zero-filled fp32 output | Pass |
| Duplicate indices | duplicate sparse row list | selected row value remains deterministic | Pass |
| Unordered indices | unordered sparse row list | selected rows match dense results | Pass |
| Inactive rows | sentinel-filled output | inactive rows preserve sentinel | Pass |
| Invalid token ids | negative and `>= vocab_size` ids | invalid positions zero-fill, valid positions match reference | Pass |
| Bad token shape | token ids shape does not match logits leading shape | raises | Pass |
| Bad output shape | output shape mismatch | raises | Pass |
| Unsupported output dtype | integer output tensor | raises | Pass |
| Bad row index dtype at C++ boundary | `int32` row indices passed directly to `_C` | raises `row_indices must be int64` | Pass |
| Out-of-range row indices | negative and too-large row ids mixed with valid id | valid row is written, invalid rows do not overwrite output | Pass |

### Numerical-Stability Matrix

| Logit Pattern | Vocab | Token Targets | Assertion | Result |
| all `0.0` | `4099` | first token | finite and matches reference | Pass |
| all `80.0` | `4099` | last token | finite and matches reference | Pass |
| all `-80.0` | `4099` | middle token | finite and matches reference | Pass |
| linearly increasing `[-80, 80]` | `4099` | last token | finite and matches reference | Pass |
| linearly decreasing `[80, -80]` | `4099` | first token | finite and matches reference | Pass |

## Validation Results

| Command / Check | Result |
| --- | --- |
| CUDA extension rebuild with CUDA 11.8 / VS BuildTools on Windows | Pass |
| Direct cp39 CUDA production matrix | `bucketed_deterministic_logp_cuda_production_matrix_passed` |
| Direct cp39 CUDA error contract | `deterministic_logp_cuda_error_contract_passed` |
| `py -3.13 -m pytest rl_engine/tests/test_dispatch.py tests\test_grpo_single_gpu_example.py tests\test_vllm_rollout_sampler.py tests\test_deepspeed_training_worker.py tests\test_deterministic_logp.py tests\test_op_accuracy.py -q` | `51 passed, 73 skipped` |
| `py -3.13 -m pytest tests\test_deterministic_logp.py -q` | `1 passed, 73 skipped` on py3.13, because CUDA extension is cp39 in this local env |
| `py -3.13 -m pre_commit run --all-files` | Pass |
| `py -3.13 -m mypy --ignore-missing-imports rl_engine/` | Pass |
| `py -3.13 -m compileall rl_engine tests\test_deterministic_logp.py` | Pass |
| `py -3.13 -m mkdocs build --strict -f mkdocs.yaml` | Pass |
| `git diff --check` | Pass, with CRLF conversion warnings only |
| `python benchmarks\benchmark_rl_kernels.py --candidate deterministic_fp32 --smoke --warmup 1 --repeat 2` | Pass |
| `python benchmarks\benchmark_rl_kernels.py --candidate deterministic_indexed_fp32 --smoke --warmup 1 --repeat 2` | Pass |
| `python benchmarks\benchmark_rl_kernels.py --candidate deterministic_fp32 --vocab-sizes 4097 --completion-lens 4 --g-sizes 2 --num-prompts 1 --mask-densities 1.0 --warmup 1 --repeat 2` | Pass |
| `python benchmarks\benchmark_rl_kernels.py --candidate deterministic_indexed_fp32 --vocab-sizes 4097 --completion-lens 4 --g-sizes 2 --num-prompts 1 --mask-densities 1.0 --warmup 1 --repeat 2` | Pass |

## Release Notes

* **New Features**
  * Added a deterministic log-probability backend with selectable `logp_backend` and optional batch-invariant enforcement (`require_batch_invariant_logp`).
  * Introduced deterministic candidates for benchmarking, plus backend selection support in the GRPO single-GPU example and rollout executor.

* **Bug Fixes**
  * Fixed DeepSpeed training config initialization to run parent validation logic.

* **Tests**
  * Added a CUDA-gated deterministic LogP test suite (correctness, bitwise repeatability, indexing semantics, and edge cases).
  * Added unit tests for configuration validation and backend alias handling.

### DIFF
--- a/csrc/deterministic_logp_kernel.cu
+++ b/csrc/deterministic_logp_kernel.cu
@@ -7,18 +7,29 @@
 
 namespace {
 
-constexpr int kDeterministicLogpBlockSize = 256;
+constexpr int kDeterministicLogpSmallBlockSize = 128;
+constexpr int kDeterministicLogpMediumBlockSize = 256;
+constexpr int kDeterministicLogpLargeBlockSize = 512;
+constexpr int kDeterministicLogpSmallVocabLimit = 128;
+constexpr int kDeterministicLogpMediumVocabLimit = 4096;
 constexpr int kDeterministicLogpWarpSize = 32;
-constexpr int kDeterministicLogpWarpCount =
-    kDeterministicLogpBlockSize / kDeterministicLogpWarpSize;
 constexpr float kDeterministicLogpNegInf = -3.4028234663852886e38F;
 
 template <int BlockSize>
-__device__ __forceinline__ float deterministicBlockReduceMax(float val) {
+struct DeterministicLogpBlockTraits {
     static_assert(
-        BlockSize == kDeterministicLogpBlockSize,
-        "deterministic logp reduction topology requires BlockSize=256");
-    __shared__ float shared[kDeterministicLogpWarpCount];
+        BlockSize == kDeterministicLogpSmallBlockSize ||
+            BlockSize == kDeterministicLogpMediumBlockSize ||
+            BlockSize == kDeterministicLogpLargeBlockSize,
+        "deterministic logp reduction topology requires a supported fixed block size");
+    static_assert(BlockSize % kDeterministicLogpWarpSize == 0, "block size must be warp-aligned");
+    static constexpr int WarpCount = BlockSize / kDeterministicLogpWarpSize;
+};
+
+template <int BlockSize>
+__device__ __forceinline__ float deterministicBlockReduceMax(float val) {
+    constexpr int WarpCount = DeterministicLogpBlockTraits<BlockSize>::WarpCount;
+    __shared__ float shared[WarpCount];
 
     int lane = threadIdx.x & (kDeterministicLogpWarpSize - 1);
     int wid = threadIdx.x / kDeterministicLogpWarpSize;
@@ -33,7 +44,7 @@ __device__ __forceinline__ float deterministicBlockReduceMax(float val) {
     }
     __syncthreads();
 
-    val = threadIdx.x < kDeterministicLogpWarpCount ? shared[lane] : kDeterministicLogpNegInf;
+    val = threadIdx.x < WarpCount ? shared[lane] : kDeterministicLogpNegInf;
     if (wid == 0) {
 #pragma unroll
         for (int offset = 16; offset > 0; offset >>= 1) {
@@ -45,10 +56,8 @@ __device__ __forceinline__ float deterministicBlockReduceMax(float val) {
 
 template <int BlockSize>
 __device__ __forceinline__ float deterministicBlockReduceSum(float val) {
-    static_assert(
-        BlockSize == kDeterministicLogpBlockSize,
-        "deterministic logp reduction topology requires BlockSize=256");
-    __shared__ float shared[kDeterministicLogpWarpCount];
+    constexpr int WarpCount = DeterministicLogpBlockTraits<BlockSize>::WarpCount;
+    __shared__ float shared[WarpCount];
 
     int lane = threadIdx.x & (kDeterministicLogpWarpSize - 1);
     int wid = threadIdx.x / kDeterministicLogpWarpSize;
@@ -63,7 +72,7 @@ __device__ __forceinline__ float deterministicBlockReduceSum(float val) {
     }
     __syncthreads();
 
-    val = threadIdx.x < kDeterministicLogpWarpCount ? shared[lane] : 0.0f;
+    val = threadIdx.x < WarpCount ? shared[lane] : 0.0f;
     if (wid == 0) {
 #pragma unroll
         for (int offset = 16; offset > 0; offset >>= 1) {
@@ -202,20 +211,56 @@ void launch_deterministic_logp_kernel(
                 "deterministic_logp_output_kernel",
                 ([&] {
                     using output_t = scalar_t;
-                    deterministic_logp_forward_kernel<
-                        input_t,
-                        output_t,
-                        kDeterministicLogpBlockSize><<<
-                        static_cast<int>(launch_rows),
-                        kDeterministicLogpBlockSize,
-                        0,
-                        at::cuda::getCurrentCUDAStream()>>>(
-                        logits.data_ptr<input_t>(),
-                        token_ids.data_ptr<int64_t>(),
-                        output.data_ptr<output_t>(),
-                        row_indices_ptr,
-                        total_rows,
-                        static_cast<int>(vocab_size));
+                    const int vocab_size_i32 = static_cast<int>(vocab_size);
+                    const int launch_rows_i32 = static_cast<int>(launch_rows);
+                    cudaStream_t stream = at::cuda::getCurrentCUDAStream();
+
+                    if (vocab_size <= kDeterministicLogpSmallVocabLimit) {
+                        deterministic_logp_forward_kernel<
+                            input_t,
+                            output_t,
+                            kDeterministicLogpSmallBlockSize><<<
+                            launch_rows_i32,
+                            kDeterministicLogpSmallBlockSize,
+                            0,
+                            stream>>>(
+                            logits.data_ptr<input_t>(),
+                            token_ids.data_ptr<int64_t>(),
+                            output.data_ptr<output_t>(),
+                            row_indices_ptr,
+                            total_rows,
+                            vocab_size_i32);
+                    } else if (vocab_size <= kDeterministicLogpMediumVocabLimit) {
+                        deterministic_logp_forward_kernel<
+                            input_t,
+                            output_t,
+                            kDeterministicLogpMediumBlockSize><<<
+                            launch_rows_i32,
+                            kDeterministicLogpMediumBlockSize,
+                            0,
+                            stream>>>(
+                            logits.data_ptr<input_t>(),
+                            token_ids.data_ptr<int64_t>(),
+                            output.data_ptr<output_t>(),
+                            row_indices_ptr,
+                            total_rows,
+                            vocab_size_i32);
+                    } else {
+                        deterministic_logp_forward_kernel<
+                            input_t,
+                            output_t,
+                            kDeterministicLogpLargeBlockSize><<<
+                            launch_rows_i32,
+                            kDeterministicLogpLargeBlockSize,
+                            0,
+                            stream>>>(
+                            logits.data_ptr<input_t>(),
+                            token_ids.data_ptr<int64_t>(),
+                            output.data_ptr<output_t>(),
+                            row_indices_ptr,
+                            total_rows,
+                            vocab_size_i32);
+                    }
                 }));
         }));
 
--- a/csrc/deterministic_logp_kernel.cu
+++ b/csrc/deterministic_logp_kernel.cu
@@ -44,7 +44,9 @@ __device__ __forceinline__ float deterministicBlockReduceMax(float val) {
     }
     __syncthreads();
 
-    val = threadIdx.x < WarpCount ? shared[lane] : kDeterministicLogpNegInf;
+    const bool has_warp_value = threadIdx.x < WarpCount;
+    const int shared_idx = has_warp_value ? threadIdx.x : 0;
+    val = has_warp_value ? shared[shared_idx] : kDeterministicLogpNegInf;
     if (wid == 0) {
 #pragma unroll
         for (int offset = 16; offset > 0; offset >>= 1) {
@@ -72,7 +74,9 @@ __device__ __forceinline__ float deterministicBlockReduceSum(float val) {
     }
     __syncthreads();
 
-    val = threadIdx.x < WarpCount ? shared[lane] : 0.0f;
+    const bool has_warp_value = threadIdx.x < WarpCount;
+    const int shared_idx = has_warp_value ? threadIdx.x : 0;
+    val = has_warp_value ? shared[shared_idx] : 0.0f;
     if (wid == 0) {
 #pragma unroll
         for (int offset = 16; offset > 0; offset >>= 1) {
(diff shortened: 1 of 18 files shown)


Flink-ddd:
Great work! Could you please add an issue to this PR?

csrc/deterministic_logp_kernel.cu
```diff
@@ -0,0 +1,342 @@
...
+    for (int offset = 16; offset > 0; offset >>= 1) {
+        val = fmaxf(val, __shfl_down_sync(0xffffffff, val, offset));
+    }
+
+    if (lane == 0) {
+        shared[wid] = val;
+    }
+    __syncthreads();
+
+    val = threadIdx.x < WarpCount ? shared[lane] : kDeterministicLogpNegInf;
```
Flink-ddd: After the per-warp shuffle, the broadcast-back line reads shared[lane]:

`val = threadIdx.x < WarpCount ? shared[lane] : kDeterministicLogpNegInf;`

shared is sized WarpCount. With BlockSize=512, WarpCount=16 but lane ranges 0–31, so threads with lane >= 16 evaluate shared[16..31], which is out of bounds. The guarding condition is in the ternary, but both branches are evaluated, so the OOB shared-memory read still happens. This triggers on any vocab_size > 4096. Please index by the guarded value:

`val = (threadIdx.x < WarpCount) ? shared[threadIdx.x] : kDeterministicLogpNegInf;`

Same fix for the sum reducer (else branch 0.0f).

csrc/deterministic_logp_kernel.cu
```diff
@@ -0,0 +1,342 @@
...
+                output.scalar_type(),
+                "deterministic_logp_output_kernel",
+                ([&] {
+                    using output_t = scalar_t;
+                    const int vocab_size_i32 = static_cast<int>(vocab_size);
+                    const int launch_rows_i32 = static_cast<int>(launch_rows);
+                    cudaStream_t stream = at::cuda::getCurrentCUDAStream();
+
+                    if (vocab_size <= kDeterministicLogpSmallVocabLimit) {
+                        deterministic_logp_forward_kernel<
```
Flink-ddd, line 225: In indexed mode multiple blocks can map to the same output[row] when row_indices contains duplicates. It's currently correct only because identical inputs produce bit-identical results, i.e. the write is idempotent. Please add a short comment documenting that duplicate/overlapping row indices are well-defined by idempotent writes, so a future change to a non-idempotent write path doesn't silently introduce a race.

rl_engine/kernels/ops/cuda/loss/logp.py
```diff
@@ -140,3 +140,84 @@ def online_indexed_fp32(
...
+    def apply_fp32(self, logits: torch.Tensor, token_ids: torch.Tensor) -> torch.Tensor:
+        logits_2d, token_ids_1d, orig_shape = self._prepare_inputs(logits, token_ids)
+        results = self._backend.deterministic_logp_forward_fp32(logits_2d, token_ids_1d)
+        return results.view(orig_shape)
+
+    def out(
+        self, logits: torch.Tensor, token_ids: torch.Tensor, output: torch.Tensor
+    ) -> torch.Tensor:
+        logits_2d, token_ids_1d, orig_shape = self._prepare_inputs(logits, token_ids)
+        output_1d = self._prepare_output(output, orig_shape)
```
Flink-ddd, line 180: test_..._rejects_bad_shapes_and_output_dtype_cuda expects ValueError(match="output shape") on shape mismatch:

```
with pytest.raises(ValueError, match="output shape"):
    op.out(logits, token_ids, torch.empty(2, 2, ...))
```

That error must be raised in _prepare_output. Please confirm _prepare_output validates output.shape == orig_shape and raises a ValueError containing "output shape" — otherwise this test will fail. (Not visible in the diff.)

examples/grpo_single_gpu.py
```diff
@@ -87,19 +101,27 @@ def select_device(requested: str) -> torch.device:
...
 
     from rl_engine.kernels.registry import kernel_registry
 
-    return kernel_registry.get_op("logp")
+    return kernel_registry.get_op(op_type)
 
 
 def is_fused_logp_backend(backend_name: str) -> bool:
-    return backend_name.startswith("FusedLogp")
+    return backend_name.startswith("FusedLogp") or backend_name == "DeterministicLogpCUDAOp"
```
Flink-ddd: String-name matching is brittle — renaming the class or adding a new deterministic variant (e.g. an SM90 deterministic op) will silently break this. Prefer a capability flag on the op:

```
class DeterministicLogpCUDAOp(FusedLogpGenericOp):
    is_batch_invariant = True

def is_fused_logp_backend(op) -> bool:
    return getattr(op, "is_batch_invariant", False) or isinstance(op, FusedLogpGenericOp)
```

setup.py
```diff
@@ -57,11 +57,14 @@ def get_extensions():
...
             "csrc/ops.cpp",
             "csrc/fused_logp_kernel.cu",
+            "csrc/deterministic_logp_kernel.cu",
             "csrc/cuda/attention/prefix_shared_attention.cu",
         ]
 
         cc_major, cc_minor = torch.cuda.get_device_capability()
-        nvcc_flags = ["-O3", "--use_fast_math", "-Xfatbin", "-compress-all"]
+        nvcc_flags = ["-O3", "-Xfatbin", "-compress-all"]
+        if os.environ.get("KERNEL_ALIGN_USE_FAST_MATH") == "1":
```
Flink-ddd: This scatters raw env-var string parsing across the codebase (same pattern repeats for KERNEL_ALIGN_NCU_LINEINFO, KERNEL_ALIGN_ALLOW_UNSUPPORTED_MSVC). I'd suggest centralizing these into an envs.py module like vLLM does, so every flag has one documented definition, consistent parsing, and a single import surface for call sites.

Create rl_engine/envs.py (or top-level envs.py next to setup.py):


```
# rl_engine/envs.py
import os
from typing import TYPE_CHECKING, Any, Callable

if TYPE_CHECKING:
    KERNEL_ALIGN_USE_FAST_MATH: bool = False
    KERNEL_ALIGN_NCU_LINEINFO: bool = False
    KERNEL_ALIGN_ALLOW_UNSUPPORTED_MSVC: bool = False

environment_variables: dict[str, Callable[[], Any]] = {
    # Opt-in --use_fast_math for CUDA kernels. Off by default because it
    # breaks bit-for-bit reproducibility required by the deterministic logp path.
    "KERNEL_ALIGN_USE_FAST_MATH": lambda: bool(
        int(os.getenv("KERNEL_ALIGN_USE_FAST_MATH", "0"))
    ),
    "KERNEL_ALIGN_NCU_LINEINFO": lambda: bool(
        int(os.getenv("KERNEL_ALIGN_NCU_LINEINFO", "0"))
    ),
    # Windows-only escape hatch for unsupported MSVC/CUDA combinations.
    "KERNEL_ALIGN_ALLOW_UNSUPPORTED_MSVC": lambda: bool(
        int(os.getenv("KERNEL_ALIGN_ALLOW_UNSUPPORTED_MSVC", "0"))
    ),
}

def __getattr__(name: str):
    if name in environment_variables:
        return environment_variables[name]()
    raise AttributeError(f"module {__name__!r} has no attribute {name!r}")

def __dir__():
    return list(environment_variables.keys())
```

Then call sites read it the same way vLLM does:



```
from rl_engine import envs

nvcc_flags = ["-O3", "-Xfatbin", "-compress-all"]
if envs.KERNEL_ALIGN_USE_FAST_MATH:
    nvcc_flags.append("--use_fast_math")
```

setup.py
```diff
@@ -106,6 +109,9 @@ def get_extensions():
         )
         if os.environ.get("KERNEL_ALIGN_NCU_LINEINFO") == "1":
             nvcc_flags.append("-lineinfo")
+        if os.name == "nt" and os.environ.get("KERNEL_ALIGN_ALLOW_UNSUPPORTED_MSVC") == "1":
```
Flink-ddd: same

```
if os.name == "nt" and envs.KERNEL_ALIGN_ALLOW_UNSUPPORTED_MSVC:
    nvcc_flags.append("-allow-unsupported-compiler")
    nvcc_flags.append("-D_ALLOW_COMPILER_AND_STL_VERSION_MISMATCH")
```


One caveat: setup.py runs before the package is importable, so either keep envs.py import-safe with zero heavy deps (as above), or place it at repo root and import it directly in setup.py.

Flink-ddd (changes requested):
Solid, well-tested PR, Happy to re-review once the shared-memory fix lands.

inaniloquentee:
> Great work! Could you please add an issue to this PR?

Sure! The issue is here: https://github.com/RL-Align/RL-Kernel/issues/96

inaniloquentee:
> Solid, well-tested PR, Happy to re-review once the shared-memory fix lands.

Good catch. The previous reducer used shared[lane], which is unsafe when WarpCount < 32.

I fixed both deterministic reducers to read from shared memory only when threadIdx.x < WarpCount; otherwise the lane uses the reduction identity value. I also added a regression/source-contract test to prevent this pattern from coming back.

Flink-ddd:
please resolve conflicts first, Thanks.

inaniloquentee:
> please resolve conflicts first, Thanks.

Conflicts resolved. Thank you for your review.

Flink-ddd (approved):
LGTM, this PR as first version still useful for furture. cc @KJLdefeated PTAL.

KJLdefeated (approved):
LGTM. Pulled the branch and tested end-to-end on an H100 (sm_90, CUDA 12.4). Builds clean with the aligned toolchain, and every claim in the PR holds up. 
Non-blocking follow-ups (fine as separate PRs):
- Perf opportunity: on H100, bs=1024 may be ~25–28% faster than 512 for 8k–64k vocab (the common LLM range) due to L2 reuse; negligible above 128k, and worse for the ≤4096 buckets. Worth a mid-bucket split later, but note it changes the reduction tree, so it's a versioned bit change, not a silent tweak.

Issue #39, opened 2026-05-31
[FEAT][kernels]: integrate Composable Kernel / ROCm FlashAttention for AMD parity


Flink-ddd:
### Context:
Following the hardware abstraction layer refactoring (Issue #36), we need to ensure our AMD users have parity with CUDA users for attention operations. We should integrate AMD's Composable Kernel or the ROCm-ported FlashAttention as the primary backend for AMD GPUs.

## Tasks:
1. Add ROCm backend dispatch logic in rl_engine/kernels/ops/rocm/attention/.
2. Wrap ROCm Flash-Attention-2 (from flash-attn ROCm fork) or natively call CK attention operators.
3. Ensure the input/output signature perfectly matches the CUDA flash_attn.py wrapper for seamless higher-level routing.
4. Update the CI/CD pipeline to skip ROCm specific tests if AMD hardware is not detected, but ensure linting passes.

FED4:
Hi, I’d like to take this. I have AMD/NVIDIA GPU access and can implement ROCm attention backend dispatch, validate parity with the CUDA wrapper, and add correctness/perf results. Please assign it to me.

Flink-ddd:
That's great, This task is primarily for AMD ROCm, so you have an AMD MI300X or 355, right?

FED4:
> That's great, This task is primarily for AMD ROCm, so you have an AMD MI300X or 355, right?

Yes, I can access MI300X.

Flink-ddd:
okay, I will assign this task to you. Thanks.

FED4:
I’ll implement the ROCm attention wrapper with the same public call signature as the existing CUDA FlashAttentionOp.

My current understanding:
- Dispatch selection lives in rl_engine/kernels/registry.py.
- The ROCm attention backend should live under rl_engine/kernels/ops/rocm/attention/.
- I’ll start with ROCm FlashAttention-2 because it is closest to the existing CUDA FlashAttention wrapper; CK can be added as a follow-up backend if needed.

[...]

PR #104, opened 2026-06-13
[FEAT][kernels]: add ROCm FlashAttention backend

FED4:
## Summary

Related to #39.

This PR adds ROCm attention support and makes external ROCm FlashAttention the default ROCm attention backend to match the issue contract. PyTorch SDPA remains available as fallback / explicit opt-out:

```bash
export RL_KERNEL_ROCM_ATTN_BACKEND=sdpa
```

The SDPA/AOTriton vs external ROCm FlashAttention performance gap is real, but I am treating that as follow-up work rather than changing the default in this PR.

## Changes

- Add `RocmFlashAttentionOp` for external FlashAttention 2 Triton AMD.
- Use `ROCM_FLASH_ATTN` as the default ROCm `attn` backend.
- Add `NativeAttentionOp` as the PyTorch SDPA fallback for FlashAttention-layout tensors.
- Add `RL_KERNEL_ROCM_ATTN_BACKEND=sdpa` opt-out for PyTorch SDPA dispatch.
- Add GQA/MQA handling to `NativeAttentionOp` for `q_heads > kv_heads`.
- Add CUDA/ROCm/native attention correctness tests against PyTorch SDPA math backend.
- Add ROCm FlashAttention environment checker and installation notes.

## ROCm backend note

This PR does not implement RL-Kernel's generic `ROCM_CK` backend.

For FlashAttention itself, this PR uses FlashAttention 2 Triton AMD. I also tried FlashAttention 2's CK backend, but its standalone build failed on the MI300/gfx942 machine before RL-Kernel was involved, so this PR does not depend on that path.

## Test environment

```text
| Item | Value |
|---|---|
| GPU | AMD Radeon Graphics, gfx942 / MI300-class |
| PyTorch | 2.9.1+rocm6.3 |
| HIP | 6.3.42134-a9a80e791 |
| PyTorch ROCm FA backend | AOTriton |
| External FlashAttention backend | Triton AMD |
```

## Correctness

Command:

```bash
FLASH_ATTENTION_TRITON_AMD_ENABLE=TRUE PYTEST_DISABLE_PLUGIN_AUTOLOAD=1 python -m pytest tests/test_attention_correctness.py -q -rs
```

Result:

```text
| Test | Result |
|---|---|
| NativeAttentionOp + ROCm FlashAttention correctness | 87 passed |
| CUDA external FlashAttention cases on ROCm machine | 40 skipped |
| Skip reason | current torch build is not CUDA platform |
```

Observed external ROCm FlashAttention diff vs PyTorch SDPA math backend:

```text
| dtype | max abs diff observed | tolerance |
|---|---:|---:|
| fp16 | ~1.01e-3 | atol=1e-3, rtol=1e-3 |
| bf16 | ~8.84e-3 | atol=2e-2, rtol=2e-2 |
```

Other checks:

```text
| Check | Result |
|---|---|
| python scripts/check_rocm_env.py | passed |
| pytest tests/test_kernel_registry.py rl_engine/tests/test_dispatch.py tests/test_attention_correctness.py -q -rs | 102 passed, 40 skipped |
| ruff check changed files | passed |
| python -m mkdocs build --strict -f mkdocs.yaml | passed |
```

## One-off benchmark

This is a quick sanity benchmark, not a formal performance claim. It compares PyTorch SDPA/AOTriton with external `RocmFlashAttentionOp`.

With `FLASH_ATTENTION_TRITON_AMD_AUTOTUNE=TRUE`:

```text
| dtype | shape (B,S,H,D) | causal | NativeAttentionOp ms | RocmFlashAttentionOp autotune ms | native/external |
|---|---:|---:|---:|---:|---:|
| fp16 | (1,128,4,64) | false | 0.020 | 0.101 | 0.19x |
| fp16 | (2,256,8,128) | false | 0.047 | 0.103 | 0.45x |
| fp16 | (1,512,8,256) | false | 0.099 | 0.116 | 0.85x |
| fp16 | (1,1024,16,64) | false | 0.118 | 0.111 | 1.07x |
| fp16 | (1,2048,16,64) | false | 0.311 | 0.344 | 0.90x |
| bf16 | (1,512,8,256) | false | 0.100 | 0.114 | 0.87x |
| bf16 | (1,1024,16,64) | false | 0.136 | 0.116 | 1.17x |
| bf16 | (1,2048,16,64) | false | 0.345 | 0.377 | 0.92x |
```

External FlashAttention wins on some cases, but not broadly enough to claim a general performance win. Further performance work can investigate CK or shape-specific routing in a separate PR.

## Release Notes

* **New Features**
  * Added ROCm attention backend support, including a configurable attention-kernel selection via `RL_KERNEL_ROCM_ATTN_BACKEND`.
  * Added a PyTorch SDPA-based attention fallback with support for GQA/MQA head layouts.

* **Documentation**
  * Updated ROCm installation guide with build/verification steps and backend switching instructions.

* **Tests**
  * Expanded attention correctness coverage (CUDA/ROCm, causal/non-causal, softmax scaling) and added negative validation cases.
  * Added ROCm dispatch/priority tests for environment-driven kernel selection.

* **Chores**
  * Improved CI token permissions and added an additional targeted pytest run.

### DIFF
--- a/rl_engine/kernels/ops/pytorch/attention/__init__.py
+++ b/rl_engine/kernels/ops/pytorch/attention/__init__.py
@@ -1,0 +1,34 @@
+# SPDX-License-Identifier: Apache-2.0
+# Copyright (c) 2026 RL-Kernel Contributors
+
+import torch
+import torch.nn.functional as F
+
+
+class NativeAttentionOp:
+    """PyTorch SDPA fallback for FlashAttention-layout tensors."""
+
+    def __call__(
+        self,
+        q: torch.Tensor,
+        k: torch.Tensor,
+        v: torch.Tensor,
+        dropout_p: float = 0.0,
+        softmax_scale: float | None = None,
+        causal: bool = False,
+    ) -> torch.Tensor:
+        q_ref = q.transpose(1, 2)
+        k_ref = k.transpose(1, 2)
+        v_ref = v.transpose(1, 2)
+        out = F.scaled_dot_product_attention(
+            q_ref,
+            k_ref,
+            v_ref,
+            dropout_p=dropout_p,
+            is_causal=causal,
+            scale=softmax_scale,
+        )
+        return out.transpose(1, 2)
+
+
+__all__ = ["NativeAttentionOp"]
--- a/rl_engine/kernels/ops/pytorch/attention/__init__.py
+++ b/rl_engine/kernels/ops/pytorch/attention/__init__.py
@@ -17,9 +17,24 @@ class NativeAttentionOp:
         softmax_scale: float | None = None,
         causal: bool = False,
     ) -> torch.Tensor:
+        # Convert FlashAttention layout to PyTorch SDPA layout:
+        # (batch, seqlen, nheads, headdim) -> (batch, nheads, seqlen, headdim)
         q_ref = q.transpose(1, 2)
         k_ref = k.transpose(1, 2)
         v_ref = v.transpose(1, 2)
+
+        q_head_num = q_ref.shape[1]
+        k_head_num = k_ref.shape[1]
+        if k_head_num != v_ref.shape[1]:
+            raise ValueError("k and v must have the same number of heads")
+
+        if q_head_num != k_head_num:
+            if q_head_num % k_head_num != 0:
+                raise ValueError("q heads must be divisible by k/v heads for GQA/MQA")
+            repeat = q_head_num // k_head_num
+            k_ref = k_ref.repeat_interleave(repeat, dim=1)
+            v_ref = v_ref.repeat_interleave(repeat, dim=1)
+
         out = F.scaled_dot_product_attention(
             q_ref,
             k_ref,
--- a/rl_engine/kernels/ops/rocm/attention/flash_attn.py
+++ b/rl_engine/kernels/ops/rocm/attention/flash_attn.py
@@ -1,0 +1,75 @@
+# SPDX-License-Identifier: Apache-2.0
+# Copyright (c) 2026 RL-Kernel Contributors
+
+import importlib.util
+import os
+
+import torch
+
+from rl_engine.utils.logger import logger
+
+
+def _select_flash_attn_backend() -> str:
+    """Select the installed FlashAttention ROCm backend before importing flash_attn."""
+    if os.environ.get("FLASH_ATTENTION_TRITON_AMD_ENABLE", "").upper() == "TRUE":
+        return "triton"
+    if importlib.util.find_spec("flash_attn_2_cuda") is None:
+        os.environ["FLASH_ATTENTION_TRITON_AMD_ENABLE"] = "TRUE"
+        return "triton"
+    return "ck"
+
+
+class RocmFlashAttentionOp:
+    """
+    Standard FlashAttention wrapper for ROCm.
+    Demonstrates the reference structure for adding new operator families.
+    """
+
+    def __init__(self):
+        if torch.version.hip is None:
+            raise RuntimeError("RocmFlashAttentionOp requires a ROCm PyTorch build.")
+
+        backend = _select_flash_attn_backend()
+        try:
+            from flash_attn import flash_attn_func
+
+            self.op = flash_attn_func
+            logger.info("Successfully linked to external flash_attn library (%s backend).", backend)
+        except ImportError as exc:
+            raise RuntimeError(
+                "ROCm FlashAttention requires a ROCm-compatible flash-attn installation. "
+                "See docs/getting_started/installation.md#rocm-backend."
+            ) from exc
+
+    def __call__(
+        self,
+        q: torch.Tensor,
+        k: torch.Tensor,
+        v: torch.Tensor,
+        dropout_p: float = 0.0,
+        softmax_scale: float | None = None,
+        causal: bool = False,
+    ) -> torch.Tensor:
+        """
+        Standard attention forward pass.
+        Args:
+            q: (batch, seqlen, nheads, headdim)
+            k: (batch, seqlen, nheads_k, headdim)
+            v: (batch, seqlen, nheads_k, headdim)
+        """
+        valid_dtypes = (torch.float16, torch.bfloat16)
+        if (
+            q.dtype not in valid_dtypes
+            or k.dtype not in valid_dtypes
+            or v.dtype not in valid_dtypes
+        ):
+            raise TypeError("FlashAttention requires FP16 or BF16 for q/k/v")
+        # PyTorch uses the CUDA device API for both CUDA and ROCm tensors.
+        if not (q.is_cuda and k.is_cuda and v.is_cuda):
+            raise ValueError("Inputs must be on a CUDA/ROCm GPU device")
+        if not (q.device == k.device == v.device):
+            raise ValueError("q, k, and v must be on the same device")
+
+        q, k, v = q.contiguous(), k.contiguous(), v.contiguous()
+
+        return self.op(q, k, v, dropout_p=dropout_p, softmax_scale=softmax_scale, causal=causal)
--- a/rl_engine/kernels/ops/rocm/attention/flash_attn.py
+++ b/rl_engine/kernels/ops/rocm/attention/flash_attn.py
@@ -1,22 +1,18 @@
 # SPDX-License-Identifier: Apache-2.0
 # Copyright (c) 2026 RL-Kernel Contributors
 
-import importlib.util
 import os
 
 import torch
 
 from rl_engine.utils.logger import logger
 
+_MAX_TESTED_ROCM_TRITON_HEAD_DIM = 512
+
 
 def _select_flash_attn_backend() -> str:
-    """Select the installed FlashAttention ROCm backend before importing flash_attn."""
-    if os.environ.get("FLASH_ATTENTION_TRITON_AMD_ENABLE", "").upper() == "TRUE":
-        return "triton"
-    if importlib.util.find_spec("flash_attn_2_cuda") is None:
-        os.environ["FLASH_ATTENTION_TRITON_AMD_ENABLE"] = "TRUE"
-        return "triton"
-    return "ck"
+    """Select the installed FlashAttention ROCm backend."""
+    return "triton"
 
 
 class RocmFlashAttentionOp:
@@ -30,12 +26,15 @@ class RocmFlashAttentionOp:
             raise RuntimeError("RocmFlashAttentionOp requires a ROCm PyTorch build.")
 
         backend = _select_flash_attn_backend()
+        if backend == "triton":
+            # flash-attn selects the ROCm CK/Triton backend at import time.
+            os.environ["FLASH_ATTENTION_TRITON_AMD_ENABLE"] = "TRUE"
         try:
             from flash_attn import flash_attn_func
 
             self.op = flash_attn_func
             logger.info("Successfully linked to external flash_attn library (%s backend).", backend)
-        except ImportError as exc:
+        except (ImportError, OSError, RuntimeError) as exc:
             raise RuntimeError(
                 "ROCm FlashAttention requires a ROCm-compatible flash-attn installation. "
                 "See docs/getting_started/installation.md#rocm-backend."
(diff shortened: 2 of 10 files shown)


rl_engine/kernels/ops/pytorch/attention/__init__.py
```diff
@@ -0,0 +1,34 @@
...
+        k: torch.Tensor,
+        v: torch.Tensor,
+        dropout_p: float = 0.0,
+        softmax_scale: float | None = None,
+        causal: bool = False,
+    ) -> torch.Tensor:
+        q_ref = q.transpose(1, 2)
+        k_ref = k.transpose(1, 2)
+        v_ref = v.transpose(1, 2)
+        out = F.scaled_dot_product_attention(
```
inaniloquentee, line 38: This SDPA call won’t handle GQA/MQA when q_heads > kv_heads, so common LLM shapes may crash. Could we enable GQA or expand k/v, and add a nheads_k < nheads test?

FED4:
@coderabbitai resume

scripts/check_rocm_env.py
```diff
@@ -0,0 +1,138 @@
...
+        return hipcc
+
+    fallback = Path("/opt/rocm/bin/hipcc")
+    if fallback.exists():
+        return str(fallback)
+
+    return None
+
+
+def _flash_attn_backend() -> str | None:
```
Flink-ddd: The CK detection looks inverted. flash_attn_2_cuda is the CUDA-compiled extension; its presence indicates a CUDA build, not a ROCm CK path. Returning "ck" here is semantically wrong on ROCm. The same logic is duplicated in rocm/attention/flash_attn.py and has the same issue. Suggest detecting based on torch.version.hip + actual import success, or just reporting "flash_attn importable: yes/no" instead of guessing the backend.

rl_engine/kernels/ops/rocm/attention/flash_attn.py
```diff
@@ -0,0 +1,75 @@
...
+
+            self.op = flash_attn_func
+            logger.info("Successfully linked to external flash_attn library (%s backend).", backend)
+        except ImportError as exc:
+            raise RuntimeError(
+                "ROCm FlashAttention requires a ROCm-compatible flash-attn installation. "
+                "See docs/getting_started/installation.md#rocm-backend."
+            ) from exc
+
+    def __call__(
```
Flink-ddd, line 43: No head_dim validation. The issue asked for explicit supported-head-dim validation and NotImplementedError on unsupported sizes (and this was agreed in the discussion), but __call__ doesn't enforce it. Please add:

```
SUPPORTED_HEAD_DIMS = (64, 128, 256)

if q.shape[-1] not in SUPPORTED_HEAD_DIMS:
    raise NotImplementedError(
        f"RocmFlashAttentionOp supports head_dim in {SUPPORTED_HEAD_DIMS}, "
        f"got {q.shape[-1]}"
    )
```

rl_engine/kernels/ops/rocm/attention/flash_attn.py
```diff
@@ -0,0 +1,75 @@
...
+        self,
+        q: torch.Tensor,
+        k: torch.Tensor,
+        v: torch.Tensor,
+        dropout_p: float = 0.0,
+        softmax_scale: float | None = None,
```
Flink-ddd, line 49: softmax_scale=None is passed straight through and relies on flash_attn_func's internal 1/sqrt(d) default matching the reference's default. The correctness test only passes because both defaults happen to agree; nothing enforces it. Suggest normalizing explicitly so behavior is independent of the downstream library's default:

```
if softmax_scale is None:
    softmax_scale = q.shape[-1] ** -0.5
```

rl_engine/kernels/ops/rocm/attention/flash_attn.py
```diff
@@ -0,0 +1,75 @@
...
+        if torch.version.hip is None:
+            raise RuntimeError("RocmFlashAttentionOp requires a ROCm PyTorch build.")
+
+        backend = _select_flash_attn_backend()
+        try:
+            from flash_attn import flash_attn_func
+
+            self.op = flash_attn_func
+            logger.info("Successfully linked to external flash_attn library (%s backend).", backend)
+        except ImportError as exc:
```
Flink-ddd: __init__ only catches ImportError, but the test file's AVAILABILITY_ERRORS is (ImportError, ModuleNotFoundError, OSError, RuntimeError) — flash-attn on ROCm can raise OSError (missing .so) or RuntimeError (HIP runtime) at import. These will leak through here, inconsistent with the test's own expectations.

```
except (ImportError, OSError, RuntimeError) as exc:
    raise RuntimeError(...) from exc
```

rl_engine/kernels/ops/rocm/attention/flash_attn.py
```diff
@@ -0,0 +1,75 @@
...
+
+import importlib.util
+import os
+
+import torch
+
+from rl_engine.utils.logger import logger
+
+
+def _select_flash_attn_backend() -> str:
```
Flink-ddd, line 13: This function mutates os.environ["FLASH_ATTENTION_TRITON_AMD_ENABLE"] as a side effect inside what reads like a pure selector. Also note that flash-attn reads this var at import time, so setting it after a prior import won't take effect. Please either set it before any flash_attn import in a documented startup path, or at minimum add a comment that this is an import-time dependency. Same side effect exists in check_rocm_env.py.

rl_engine/kernels/registry.py
```diff
@@ -72,24 +75,43 @@ def __init__(self):
...
                 ],
-                "attn": [OpBackend.FLASH_ATTN, OpBackend.TRITON_GENERIC, OpBackend.PYTORCH_NATIVE],
+                "attn": [OpBackend.FLASH_ATTN, OpBackend.TRITON_GENERIC, OpBackend.PYTORCH_ATTN],
                 "grpo_loss": [OpBackend.TRITON_GRPO_LOSS, OpBackend.PYTORCH_GRPO_LOSS],
                 # Default dispatch logic for new operators
             },
             "rocm": {
                 "logp": [OpBackend.ROCM_AITER, OpBackend.TRITON_GENERIC, OpBackend.PYTORCH_NATIVE],
-                "attn": [OpBackend.TRITON_GENERIC, OpBackend.PYTORCH_NATIVE],
+                "attn": [
```
Flink-ddd, line 89: This makes PYTORCH_ATTN (SDPA) the default ROCm attention backend, with ROCM_FLASH_ATTN demoted to opt-in. Issue #39 asked for ROCm FlashAttention as the primary backend. The benchmark rationale is reasonable, but the PR description itself labels it "not a formal performance claim," and this requirement reversal wasn't agreed in the issue thread. Can you sync this decision back to #39 and get maintainer sign-off before we land the default change? The code is fine; this is a scope/contract question.

rl_engine/kernels/registry.py
```diff
@@ -72,24 +75,43 @@ def __init__(self):
...
-                "attn": [OpBackend.PYTORCH_NATIVE],
+                "attn": [OpBackend.PYTORCH_ATTN],
                 "grpo_loss": [OpBackend.PYTORCH_GRPO_LOSS],
             },
         }
         logger.info(f"KernelRegistry initialized for {device_ctx.device_type}")
+        self._adjust_priority_from_env()
         self._adjust_priority_for_hardware()
 
+    def _adjust_priority_from_env(self):
```
Flink-ddd, line 108: _adjust_priority_from_env() runs before _adjust_priority_for_hardware(). If the hardware pass also touches rocm/attn, an explicit env opt-in could be silently overridden by the hardware default. Explicit env should generally win — please confirm ordering and add a test covering env + hardware together.

rl_engine/tests/test_dispatch.py
```diff
@@ -25,6 +25,23 @@ def test_device_and_registry():
...
+def test_rocm_attention_uses_native_sdpa_by_default(monkeypatch):
+    monkeypatch.delenv("RL_KERNEL_ROCM_ATTN_BACKEND", raising=False)
+
+    registry = KernelRegistry()
+
+    assert registry._priority_map["rocm"]["attn"][0] == OpBackend.PYTORCH_ATTN
+
+
+def test_rocm_attention_flash_attn_opt_in(monkeypatch):
+    monkeypatch.setenv("RL_KERNEL_ROCM_ATTN_BACKEND", " flash_attn ")
```
Flink-ddd: Nice catch testing .strip() with " flash_attn ". Missing coverage for the other branches that registry.py actually implements: case-insensitivity, the unknown-value warning path, and explicit "native"/"sdpa". Suggest:

```
@pytest.mark.parametrize("val", ["FLASH_ATTN", "flash-attn", "Flash_Attention"])
def test_rocm_attn_flash_opt_in_case_insensitive(monkeypatch, val):
    monkeypatch.setenv("RL_KERNEL_ROCM_ATTN_BACKEND", val)
    registry = KernelRegistry()
    assert registry._priority_map["rocm"]["attn"][0] == OpBackend.ROCM_FLASH_ATTN
```

tests/test_attention_correctness.py
```diff
@@ -0,0 +1,351 @@
...
+        return
+
+    raise RuntimeError("PyTorch SDPA backend selector is unavailable")
+
+
+def should_print_attention_diff():
+    return os.getenv("PRINT_ATTENTION_DIFF", "").lower() in {"1", "true", "yes"}
+
+
+def pytorch_sdpa_reference(q, k, v, *, causal, softmax_scale):
```
Flink-ddd, line 57: The reference is computed in fp32 and compared against the low-precision op cast up to float. That validates "low-precision op vs fp32 golden value," which is correct, but it does not validate cross-backend (CUDA vs ROCm) numerical agreement. Suggest noting this in the docstring so future readers don't assume it guarantees backend parity.

Flink-ddd (changes requested):
Thanks for the thorough PR, Happy to re-review once the blocking items are addressed. The core wrapper and test scaffolding are in good shape

inaniloquentee (changes requested):

scripts/check_rocm_env.py
FED4: You're right.  The goal of this script is to check availibility of backends. Guessing backend is a bit redundant and I will make it simpler.

rl_engine/kernels/ops/rocm/attention/flash_attn.py
FED4: I verified this against the installed ROCm Triton flash-attn backend before adding the guard.

With FLASH_ATTENTION_TRITON_AMD_ENABLE=TRUE, runtime selection reports USE_TRITON_ROCM=True and flash_attn_gpu resolves to flash_attn.flash_attn_triton_amd.interface_fa. I tested fp16/bf16 against the fp32 SDPA reference for causal and non-causal attention. Head dims 32, 40, 59, 64, 65, 80, 96, 111, 112, 127, 128, 129, 160, 192, 224, 256, 257, 384, 511, and 512 pass; 513 fails with Triton shared-memory OutOfResources. 

So I suggest to reject head_dim > 512.

rl_engine/kernels/ops/rocm/attention/flash_attn.py
FED4: I 'll set it before import.

rl_engine/kernels/registry.py
FED4: You're right, env should be prioritized.

rl_engine/kernels/ops/pytorch/attention/__init__.py
FED4: Since enable_gqa of pytorch SDPA only works on pytorch 2.5+ while we require pytorch 2.4.1, I 'll do expand KV.

rl_engine/kernels/registry.py
FED4: I will still use rocm flash attention as default to keep everything simple in this stage.
Permformance issue can be looked into in a separate PR.

rl_engine/kernels/ops/pytorch/attention/__init__.py
inaniloquentee: > Since enable_gqa of pytorch SDPA only works on pytorch 2.5+ while we require pytorch 2.4.1, I 'll do expand KV.

Makes sense, thanks for keeping PyTorch 2.4.1 compatibility. The manual k/v LGTM.

inaniloquentee (approved):

Flink-ddd:
@FED4 please resolve CI error first, then We can merge this PR first.

Flink-ddd:
@FED4 please resolve conflicts first, once done, I will merge this PR, Thanks.

Flink-ddd (approved):
LGTM now, Thank you for your update.

PR #122, opened 2026-06-15
[FEAT][kernels] Fused linear log-prob without materializing logits — SM90 TMA + tensor-core forward, chunked backward

KJLdefeated:
## Summary

Fused linear log-prob op (CUDA(SM90) + Triton + native reference + chunked backward landed earlier on the branch).

| Backend | Status |
| --- | --- |
| CUDA SM90 (Hopper) | TMA-streamed, double-buffered, register-blocked tensor-core forward (`ldmatrix` + `mma.sync.m16n8k16`, fp32 accum) + split-V; online softmax. |
| CUDA / ROCm (Triton) | Portable online-softmax forward + Liger-style chunked backward — the semantic baseline and tolerance target. |
| PyTorch native | `F.linear` + `log_softmax` + `gather` reference / CPU fallback. |

The SM90 op requires bf16 hidden/weight with `D % 32 == 0`; for any other input(fp32/fp16, awkward `D`, CPU) it **transparently falls back** to Triton (else native), so it stays a drop-in.

## Implementation (SM90 forward)

- **TMA** (`cp.async.bulk.tensor`, mbarrier-completed) streams `H`/`W` tiles into shared memory, **double-buffered** so the next hidden slice loads while the current one feeds the MMAs.
- **Register-blocked M-tiling** (`BM = 256` rows/CTA) so the weight matrix is re-read only `N/BM` times from HBM.
- **Split-V**: the vocab loop is partitioned across `blockIdx.y` so the grid fills all SMs even when `N/BM` alone is too few CTAs; each split emits a partial online-softmax state and a tiny combine kernel merges them (standard log-sum-exp merge). Adds only `3·n_split·N` floats of scratch — independent of `V`.

## Performance (H100, bf16, N=4096, D=2048)

**Forward latency (ms) and peak forward VRAM:**

| shape (N×D×V) | native | Triton | **SM90** | SM90 vs Triton | peak fwd VRAM (native → fused) |
| --- | --- | --- | --- | --- | --- |
| 4096×2048×32768 | 1.79 | 6.42 | **3.41** | **1.88×** | 1280 MB → **~0 MB** |
| 4096×2048×50257 | 9.96 | 9.82 | **4.88** | **2.01×** | 1965 MB → **~0 MB** |
| 4096×2048×131072 | 7.28 | 25.56 | **12.88** | **1.98×** | 5120 MB → **~0 MB** |

**Forward + backward latency (ms):**

| shape (N×D×V) | native | Triton | **SM90** |
| --- | --- | --- | --- |
| 4096×2048×32768 | 4.25 | 15.86 | 12.69 |
| 4096×2048×50257 | 23.29 | 47.20 | 42.20 |
| 4096×2048×131072 | 17.05 | 117.62 | 104.97 |

`python benchmarks/benchmark_linear_logp.py`. Forward peak memory is the per-CTA shared-memory tiles — **independent of `V`**.

## Correctness / tests

`python -m pytest tests/test_linear_logp.py` — 17/17 pass.

### Bug fix: `fused_logp_sm90.cu`
While enabling the SM90 build I found the existing TMA online-softmax kernel (`csrc/cuda/fused_logp_sm90.cu`, the materialized-logits `fused_logp`) **fails to compile** whenever the SM90 path is actually built — two latent, version-independent errors that were dormant because the prebuilt extension had no SM90 symbols (the path
had never been exercised).

1. **`CUDART_INF_F` undefined** — the kernel uses `-CUDART_INF_F` but never includes `<math_constants.h>` (not pulled in transitively).
   ```
   fused_logp_sm90.cu(62): error: identifier "CUDART_INF_F" is undefined
   ```
   Fix: `#include <math_constants.h>`.

2. **`TmaTypeTraits<c10::BFloat16>` is an incomplete type** — the host wrapper passes `logits.data_ptr<at::BFloat16>()` (i.e. `c10::BFloat16*`) straight into `init_tensor_map`, but `TmaTypeTraits` is only specialized for `nv_bfloat16`/`float`,
   so template instantiation fails.
   ```
   tma_utils.cuh: error: incomplete type "TmaTypeTraits<c10::BFloat16>" is not allowed
   ```
   Fix: `reinterpret_cast<const nv_bfloat16*>(logits.data_ptr<at::BFloat16>())`(matching how the linear kernel and the rest of the codebase bridge `at::BFloat16` → `nv_bfloat16`).

## Release Notes

* **New Features**
  * Added a CUDA SM90 fused Linear LogP backend (with optional bias) and automatic dispatch across fused, Triton, and native implementations.
  * Added a CUDA-only benchmark script to compare latency, forward+backward performance, and peak VRAM across configurable shapes (optionally including fused SM90 results when available).

* **Documentation**
  * Added Linear LogP operator documentation, including semantics, constraints, supported backends, and benchmark guidance.

* **Tests**
  * Added comprehensive correctness, gradient, fallback, error-handling, and dispatch tests covering native, Triton, and SM90 fused backends (with hardware-dependent skips).


Flink-ddd:
cc @inaniloquentee @maxiaosong1124 PTAL

rl_engine/kernels/ops/cuda/loss/linear_logp.py
```diff
@@ -0,0 +1,148 @@
...
+            raise ValueError(
+                f"hidden dim {hidden.size(-1)} must match lm_head_weight dim "
+                f"{lm_head_weight.size(-1)}"
+            )
+        # The compiled forward is bf16-only and needs D % BK == 0. For anything
+        # else (fp32/fp16 references, awkward hidden dims, CPU tensors) fall back
+        # to a portable backend so the op stays a drop-in for any input.
+        if not _sm90_supported(hidden, lm_head_weight):
+            return _fallback_op()(hidden, lm_head_weight, target_ids, bias)
+        return _FusedLinearLogpSM90Function.apply(hidden, lm_head_weight, bias, target_ids)
```
inaniloquentee, line 152: Could we add the same target/bias validation here before launching the SM90 kernel? Right now this fast path only checks the hidden dim; if target_ids has the wrong shape/length, the CUDA kernel can read target[row_base + r] out of bounds, and a bad bias shape/device can similarly become an invalid device access. Catching this as a Python/C++ validation error would keep bad inputs from turning into CUDA illegal memory access.

inaniloquentee (changes requested):

rl_engine/kernels/ops/cuda/loss/linear_logp.py
KJLdefeated: Thx for suggestion! I will add the target/bias validation before kernel launch.

KJLdefeated:
Hi @inaniloquentee, sorry for late reply, I have resolved the requests.

maxiaosong1124:
Sorry for the late reply. I fully agree with the suggestion provided by @inaniloquentee. Since you've already addressed the issue on your side, this PR is now quite solid.

Flink-ddd:
Good work! Thank you for your contribution, Could you add an issue to this PR?

inaniloquentee:
> Hi @inaniloquentee, sorry for late reply, I have resolved the requests.

Nice work! This covers the original target length and bias cases. One small remaining guard: can we also check lm_head_weight.device == hidden.device before the SM90 launch?

KJLdefeated:
Hi @Flink-ddd, @inaniloquentee and @maxiaosong1124, thx for review, the requested changes are resolved. The issue for this PR is here #155 .

inaniloquentee (approved):

inaniloquentee:
> Hi @Flink-ddd, @inaniloquentee and @maxiaosong1124, thx for review, the requested changes are resolved. The issue for this PR is here #155 .

LGTM now, thx for your contribution!

rl_engine/kernels/ops/cuda/loss/linear_logp.py
```diff
@@ -0,0 +1,165 @@
...
+            )
+        if bias is not None:
+            if bias.numel() != lm_head_weight.size(0):
+                raise ValueError(
+                    f"bias must have V={lm_head_weight.size(0)} elements, got {bias.numel()}"
+                )
+            if bias.device != hidden.device:
+                raise ValueError(
+                    f"bias device {bias.device} must match hidden device {hidden.device}"
+                )
```
Flink-ddd, line 142: We need a value-bounds check for target_ids here before launching the kernel. The current fast path only checks the shape. If target_ids contains padding tokens (-100) or out-of-bounds values, it will cause silent errors in the forward pass and wrap-around memory corruption in the backward pass.

Could you add something like this?

```python
        min_target = target_ids.min().item()
        max_target = target_ids.max().item()
        if min_target < 0 or max_target >= lm_head_weight.size(0):
            raise ValueError(
                f"target_ids contains out-of-bounds values. Expected range [0, {lm_head_weight.size(0)}-1], "
                f"got [{min_target}, {max_target}]."
            )

File: rl_engine/kernels/ops/triton/loss/linear_logp.py [hunk 3]
```diff
@@ -0,0 +1,218 @@
...
+        if hidden.shape[:-1] != target_ids.shape:
+            raise ValueError(
+                f"hidden leading shape {tuple(hidden.shape[:-1])} must match "
+                f"target_ids shape {tuple(target_ids.shape)}"
+            )
+        if lm_head_weight.size(-1) != hidden.size(-1):
+            raise ValueError(
+                f"hidden dim {hidden.size(-1)} must match lm_head_weight dim "
+                f"{lm_head_weight.size(-1)}"
+            )
```
Flink-ddd, line 192: same with rl_engine/kernels/ops/cuda/loss/linear_logp.py FusedLinearLogpSM90Op.apply

rl_engine/kernels/ops/cuda/loss/linear_logp.py
```diff
@@ -0,0 +1,165 @@
...
+        n, d = hidden_2d.shape
+        v = weight.shape[0]
+        dt = weight.dtype
+        g = grad_logp.reshape(-1).to(torch.float32)
+
+        grad_h = torch.empty_like(hidden_2d, dtype=torch.float32)
+        grad_w = torch.zeros(v, d, device=weight.device, dtype=torch.float32)
+        grad_b = torch.zeros(v, device=weight.device, dtype=torch.float32) if ctx.has_bias else None
+
+        chunk = max(1, min(n, _BWD_CHUNK_ELEMS // v))
```
Flink-ddd: This entire chunked backward logic is duplicated identically in triton/loss/linear_logp.py. To keep the codebase maintainable and maintain a Single Source of Truth, could we extract this for loop into a shared utility function? 

For example, creating a helper in a shared utils.py or base class:

```python
def compute_chunked_linear_logp_backward(hidden_2d, weight, bias_t, target_1d, g, chunk_elems):
    # Move the Liger-style chunked loop here
    # ...
    return grad_h, grad_w, grad_b

File: benchmarks/benchmark_linear_logp.py [hunk 5]
```diff
@@ -0,0 +1,190 @@
...
+    baseline = torch.cuda.memory_allocated()
+    torch.cuda.reset_peak_memory_stats()
+    for _ in range(iters):
+        fn()
+    torch.cuda.synchronize()
+    return (torch.cuda.max_memory_allocated() - baseline) / (1024**3)
+
+
+def run_benchmark(args):
+    if device_ctx.device_type != "cuda":
```
Flink-ddd: Since the Triton implementation is designed to be portable (and ROCm compatible as mentioned in the PR), this strict "cuda" check will prevent benchmarking on AMD devices. 

Let's relax this check. You can replace it with:

```python
    if device_ctx.device_type not in ["cuda", "xpu", "hip"]:
        raise RuntimeError("linear_logp benchmark requires a compatible GPU device.")

File: rl_engine/kernels/ops/cuda/loss/linear_logp.py [hunk 6]
```diff
@@ -0,0 +1,165 @@
...
+            x = hidden_2d[i0:i1]
+            logits = torch.matmul(x, weight.t())
+            if ctx.has_bias:
+                logits = logits + bias_t
+            dz = torch.softmax(logits.float(), dim=-1).neg_()
+            rows = torch.arange(i1 - i0, device=dz.device)
+            dz[rows, target_1d[i0:i1].long()] += 1.0
+            dz *= g[i0:i1].unsqueeze(1)
+
+            dz_dt = dz.to(dt)
```
Flink-ddd: Just a quick confirmation on this explicit downcast: dz is computed in fp32, and here we cast it back to dt (bf16/fp16) before the matmul. 

While this perfectly matches the Triton test precision and maximizes Tensor Core throughput, it might introduce slight quantization errors. If TF32 is enabled, doing the matmul directly in fp32 is also very fast and keeps higher gradient fidelity. 

Is this explicit downcast intentionally kept to strictly pass the cross-architecture precision tests? If yes, it's totally fine to leave it as is!

Flink-ddd (changes requested):
Thanks for the excellent PR!  It's a highly optimized piece of work that brings significant memory savings for our RL workflows. Once the boundary validation and the DRY refactoring are addressed, this will be good to merge.

rl_engine/kernels/ops/cuda/loss/linear_logp.py
KJLdefeated: I see, thx for suggetion!

rl_engine/kernels/ops/cuda/loss/linear_logp.py
KJLdefeated: Ok, I will put this in shared file

rl_engine/kernels/ops/cuda/loss/linear_logp.py
KJLdefeated: Yes, this is designed by intentional since the following matmul requires two tensors in same device.
> Is this explicit downcast intentionally kept to strictly pass the cross-architecture precision tests?

Yes, the precision checks are passed on Hopper. bf16×bf16→fp32-accum is well-defined identically on CUDA and ROCm.

KJLdefeated:
@Flink-ddd Thank you for the suggestions!!!
I have modified the following:
1. value-bounds check in `rl_engine/kernels/ops/cuda/loss/linear_logp.py` and `rl_engine/kernels/ops/triton/loss/linear_logp.py`.
2. I put the chunked backward logic in `rl_engine/kernels/ops/pytorch/loss/linear_logp.py`, so that the CUDA and Triton ops can share the backward reference. I did not create new file to increase codebase complexity.
3. `benchmarks/benchmark_linear_logp.py`: The device checks have been relax.
4. The downcast design is intentional since the following matmul requires two tensors in same device. The precision checks are passed on Hopper. bf16×bf16→fp32-accum is well-defined identically on CUDA and ROCm.

Flink-ddd:
Thank you for your updated, please resolve conflicts first. once done, I will merge this PR.

KJLdefeated:
@Flink-ddd Done.

Flink-ddd (approved):
LGTM now, Thanks.

Issue #108, opened 2026-06-14
[WS1] Ground-truth harness + numerical contract for batch-invariant ops


Flink-ddd:
Part of WS1 — Full Batch-Invariant Forward Chain (epic: #<WS1 tracking issue>)

## Why
This is the foundation for the entire batch-invariant forward chain. Every other WS1 op (RMSNorm, matmul, attention, logprob) validates against the reference and tolerance policy defined here. It must land first; nothing else in WS1 can be verified without it.

## Scope
Build a deterministic ground-truth harness and define the numerical contract that all batch-invariant ops are tested against.

- Fixed standard-Transformer model (e.g. Llama-3-8B or Qwen3-8B dense), fixed prompt set, fixed seed.
- Deterministic reference path for every op (single batch, fixed reduction order), used as the comparison target.
- A documented tolerance policy: bitwise where feasible, tight-tolerance (with an explicit threshold per dtype) otherwise.
- A reusable test utility other op-issues can import to run their batch-config sweep (batch=1/N, chunked-prefill on/off, padding-layout variations) against the reference.

## Out of scope
- The op implementations themselves (covered by their own issues).
- Multi-GPU / distributed (WS2).
- MoE, linear attention, FP8.

## Acceptance criteria
- A single command produces reference outputs for the fixed model + prompt set + seed, reproducibly.
- The tolerance policy is written down (per-dtype thresholds, bitwise vs tight-tolerance) and agreed by the WS1 owners.
- A shared test helper exists that any op-issue can call to assert "output stays within tolerance across batch configs."
- At least one existing op (e.g. logprob #96) is wired through the harness as a smoke test, proving the harness works end to end.

## Notes
- This issue blocks the other WS1 ops — assign and start it on day 1.
- PR breakdown: owner to list planned PRs below before starting.

## Planned PRs
- [ ] (owner fills in)

Issue #153, opened 2026-06-17
[WS1] Backward-pass consistency across all ops


Flink-ddd:
Part of WS1 — Full Batch-Invariant Forward Chain (epic: #<WS1 tracking issue>)

## Why

A forward-aligned chain still breaks training if gradient reductions drift with batch shape — the optimizer then sees batch-dependent gradients and the run diverges even though inference looked aligned. Backward invariance is a separate, explicit requirement, not something the forward checks cover. This issue makes "backward also invariant" a first-class, cross-op acceptance condition.

## Scope

Make batch-invariant backward a required, tested property of every WS1 op.

- Define the backward-invariance check in the #108 harness (gradient outputs compared across batch configs, same tolerance policy as forward).
- Ensure each op's gradient reduction (`dx`, `dweight`, `dW`, etc.) uses a fixed, batch-shape-independent order — no `atomicAdd` in backward.
- Cover the most reduction-heavy backward first: RMSNorm `dweight`, GEMM `dW`/`dX`, attention backward, embedding-grad scatter.
- Validate gradients across batch=1/N, chunked-prefill on/off, and padding layouts.

## Out of scope

- Re-implementing each op (each op issue owns its own backward kernel and fix; this issue owns the cross-cutting requirement, reusable gradient check, and status matrix).
- Optimizer / training-loop changes; multi-GPU gradient synchronization (WS2).
- FP8 backward.

## Acceptance criteria

- The #108 harness has a reusable gradient-invariance assertion any op-issue can call.
- Each WS1 op with a backward passes gradient-invariance across the full sweep (bitwise or within #108 tolerance).
- No op uses `atomicAdd` or shape-dependent accumulation in its backward path.
- A short matrix records, per op, both forward-pass and backward-pass invariance status; gradient-drift reports include max abs diff, relative diff, tensor name, and first failing op.

## Notes

- Depends on #108; cross-cuts every op issue (RMSNorm, matmul, attention, embedding / LM head, etc.).
- Best owned as a shared concern: this issue provides the gradient check and status matrix; each op issue is responsible for fixing its own backward path and passing the check.

## Planned PRs

- [ ] Add a reusable gradient-invariance assertion to the #108 harness
- [ ] Per-op backward test requirements (RMSNorm `dweight`, GEMM `dW`/`dX`, attention backward first)
- [ ] Gradient-diff reporting utility (max abs / relative diff, tensor name, first failing op)
[...]

PR #160, opened 2026-06-20
feat(ws1): NativeRMSNormOp pure-PyTorch ground-truth reference + numerical contract tests

maxiaosong1124:
## Summary

Adds the pure-PyTorch ground-truth reference op for **RMSNorm** (pre-norm / QK-Norm)
as the first WS1 batch-invariant operator built on top of the numerical contract
defined in #108. Ships the op, its registry wiring, and a 16-case test suite that
pins down both alignment axes (Axis-A bitwise batch invariance, Axis-B per-dtype
tolerance).

Refs #108

## Terminology

This PR uses the WS1 alignment vocabulary from #108:

- **Axis-A — batch invariance (reproducibility).** A row's output must not depend on
  how many rows share the batch (batch size, slicing, padding). Asserted **bitwise**
  (`torch.equal`). This is what keeps train-time (large batch) and sample-time
  (small batch / dynamic padding) numerics identical so the policy ratio doesn't drift.
- **Axis-B — accuracy.** The low-precision (bf16 / fp16) forward must stay within a
  documented per-dtype tolerance of the fp32 ground-truth. Asserted with `torch.allclose`
  + per-dtype thresholds.

## Motivation / Context

#108 establishes the ground-truth harness and numerical contract for the WS1
batch-invariant forward chain. RMSNorm is required on two normalized dims of the
target model (Qwen3-8B dense):

- `hidden = 4096` — input / post-attention norm
- `head_dim = 128` — QK-Norm (per-head RMSNorm on Q and K)

This PR provides the deterministic fp32 reference path those downstream kernels
(Triton / CUDA / ROCm RMSNorm) will be validated against.

## Changes

- **`rl_engine/kernels/ops/pytorch/norm/rms_norm.py`** — `NativeRMSNormOp`
  - `forward()` — accumulate in fp32, cast result back to `x.dtype` (Axis-B candidate path)
  - `forward_fp32()` — fp32 accumulation, forced fp32 output (ground-truth / backward golden source)
  - Formula: `out = x * rsqrt(mean(x^2, dim=-1) + eps) * weight`
  - `eps` lives **inside** the sqrt; **plain** weight scaling (not the `1 + weight` variant)
  - Shape guard: `weight` must be 1-D of size `x.shape[-1]`
- **`rl_engine/kernels/registry.py`** — register `PYTORCH_NATIVE_RMS_NORM`
  and add `rms_norm` dispatch to the cuda / rocm / cpu priority maps
- **`tests/test_rms_norm.py`** — 16 tests (details below)

|---|---|
| Deterministic reference path, fixed reduction order | `forward_fp32()` accumulates in fp32 along `dim=-1`; tests use fixed-seed `torch.Generator` so outputs are reproducible |
| Per-dtype tolerance policy (bitwise vs tight-tolerance) | **Axis-A** asserted **bitwise** (`torch.equal`); **Axis-B** asserted within documented per-dtype thresholds — bf16 `atol=2e-2, rtol=1.6e-2`, fp16 `atol=1e-3, rtol=1e-3` |
| Batch-config sweep / validation helper | Batch-invariance checks compute on the full batch, then assert sliced/padded rows are **bitwise identical** to their full-batch counterparts |
| Both normalized dims covered | Every correctness/invariance test is parametrized over `hidden=4096` and `head_dim=128` |

## Test Environment

| | |
|---|---|
| OS | Ubuntu 22.04.5 LTS (kernel 5.15.0-122-generic) |
| Python | 3.12.3 |
| PyTorch | 2.8.0+cu128 |
| CUDA / cuDNN | 12.8 / 9.10.02 (driver 580.65.06) |
| pytest | 9.0.3 |
| GPU | NVIDIA H20 |

## Testing

Run from the repo root with `python -m pytest` (the `-m` form puts the repo on
```bash
python -m pytest tests/test_rms_norm.py

→ 16 passed, covering:

- Correctness vs an independent hand-written fp32 formula (bitwise, both dims)
- Axis-A batch invariance: row output is independent of batch size — slice and
padding variants, asserted bitwise
- dtype paths: forward follows input dtype; forward_fp32 forces fp32
- Axis-B low-precision (bf16 / fp16) within tolerance of the fp32 reference
- eps inside sqrt (zero input → finite zero output)
- plain weight scaling (rules out the 1 + weight variant)
- shape guard fires on wrong-size / non-1-D weight
- purity (inputs not mutated in place)
- gradient flow (fp32 autograd = backward golden source)
- registry dispatch resolves rms_norm → NativeRMSNormOp

Rebased onto latest upstream/main; registry dispatch for the neighboring
ratio_kl / grpo_loss ops verified intact after conflict resolution.

Checklist

- [x] Pure-PyTorch reference, no custom extension required
- [x] Both Qwen3-8B normalized dims (4096, 128) covered
- [x] Axis-A bitwise batch invariance enforced
- [x] Axis-B per-dtype tolerance documented and tested
- [x] Registered in OpBackend + cuda/rocm/cpu priority maps
- [x] All 16 tests pass locally

---

<!-- This is an auto-generated comment: release notes by coderabbit.ai -->
## Summary by CodeRabbit

* **New Features**
  * Added a native pure-PyTorch RMSNorm operator backend and enabled `"rms_norm"` dispatch on CUDA, ROCm, and CPU.
  * `forward` now preserves input dtype (fp16/bf16/fp32), while `forward_fp32` forces fp32 outputs.
  * Enforces weight shape validation and correct `eps` behavior; output respects the requested dtype and scales linearly with weight.

* **Tests**
  * Added extensive pytest coverage for correctness vs fp32 references, dtype routing, error handling, immutability, gradient sanity, batch-slice invariance (forward and backward), and registry dispatch.
<!-- end of auto-generated comment: release notes by coderabbit.ai -->

### CODE REVIEW

File: tests/test_rms_norm.py [hunk 1]
```diff
@@ -0,0 +1,130 @@
...
+
+
+# 9. Gradient flows (fp32 autograd = backward golden source)
+def test_gradient_flows():
+    op = NativeRMSNormOp()
+    x = _rand((2, _HIDDEN), seed=13).requires_grad_(True)
+    w = _rand((_HIDDEN,), seed=14).requires_grad_(True)
+    op.forward_fp32(x, w).sum().backward()
+    assert torch.isfinite(x.grad).all() and torch.isfinite(w.grad).all()
+
```
Flink-ddd, line 138: You have thoroughly covered forward Axis-A (Batch Invariance), but WS1 requires gradients to be batch-invariant as well (vital for #153). Currently, test_gradient_flows only checks if gradients exist and are finite. You need to prove that slicing a batch yields identical gradients to computing the full batch.

Code Example:

```
# Add this new test case
def test_backward_batch_invariance_slice():
    """Axis A: Gradients must be bitwise identical regardless of batch size."""
    op = NativeRMSNormOp()
    
    # Full batch forward & backward
    w_full = _rand((_HIDDEN,), seed=1).requires_grad_(True)
    x_full = _rand((8, 32, _HIDDEN), seed=2).requires_grad_(True)
    out_full = op.forward_fp32(x_full, w_full)
    
    dy_full = _rand(out_full.shape, seed=3)
    out_full.backward(dy_full)
    grad_x_full_sliced = x_full.grad[:1].clone()
    
    # Sliced batch forward & backward (batch size = 1)
    w_slice = _rand((_HIDDEN,), seed=1).requires_grad_(True)
    x_slice = _rand((8, 32, _HIDDEN), seed=2)[:1].detach().requires_grad_(True)
    out_slice = op.forward_fp32(x_slice, w_slice)
    
    out_slice.backward(dy_full[:1]) # Use matching slice of upstream grad
    
    # Assert x.grad is bitwise identical
    assert torch.equal(x_slice.grad, grad_x_full_sliced)
```

rl_engine/kernels/ops/pytorch/norm/rms_norm.py
```diff
@@ -0,0 +1,70 @@
...
+        eps: float,
+        output_dtype: torch.dtype,
+    ) -> torch.Tensor:
+        if weight.dim() != 1 or weight.shape[0] != x.shape[-1]:
+            raise ValueError(
+                f"weight must be 1-D of size x.shape[-1]={x.shape[-1]}, "
+                f"got tuple(weight.shape)={tuple(weight.shape)}"
+            )
+        x_f = x.float()
+        var = x_f.pow(2).mean(dim=-1, keepdim=True)
```
Flink-ddd, line 67: The line var = x_f.pow(2).mean(...) explicitly materializes a full-sized FP32 tensor in memory before reducing. For a reference/ground-truth operator, code readability is more important than VRAM efficiency, so this is perfectly fine here. However, keep this memory spike in mind when reviewing the downstream Triton/CUDA implementations for long-context workloads. No code change is strictly required for this PR.

Flink-ddd (changes requested):
This is a solid, production-ready PR that strictly adheres to the #108 numerical contract. Here are a few professional suggestions for refinement before merging:

tests/test_rms_norm.py
```diff
@@ -0,0 +1,130 @@
...
+    var = x_f.pow(2).mean(dim=-1, keepdim=True)
+    return x_f * torch.rsqrt(var + eps) * weight.float()
+
+
+# 1. Correctness vs an independent fp32 formula (both normalized dims)
+@pytest.mark.parametrize("N", [_HIDDEN, _HEAD_DIM])
+def test_forward_fp32_matches_manual_reference(N):
+    op = NativeRMSNormOp()
+    x, w = _rand((2, 16, N), seed=0), _rand((N,), seed=1)
+    assert torch.equal(op.forward_fp32(x, w), _manual_rms_norm(x, w))
```
KJLdefeated, line 49: `_manual_rms_norm` is the same expression as `_rms_norm` in the same float order, which is why torch.equal passes bitwise. The downside is that if there is a bug in the formula (eps placement, wrong reduction dim) would be present in both sides and still go green. 
Note it has to be allclose, not torch.equal since `F.rms_norm` may reduce in a different order, and that non-identical reduction is exactly what makes it a real check:
```python
import torch.nn.functional as F

@pytest.mark.parametrize("N", [_HIDDEN, _HEAD_DIM])
def test_forward_fp32_matches_torch_reference(N):
    op = NativeRMSNormOp()
    x, w = _rand((2, 16, N), seed=0), _rand((N,), seed=1)
    ref = F.rms_norm(x.float(), (N,), weight=w.float(), eps=_EPS)
    torch.testing.assert_close(op.forward_fp32(x, w), ref, rtol=1e-6, atol=1e-6)
```
Fine to keep the hand-written reference as a secondary sanity check, but it shouldn't be the primary correctness assertion.

KJLdefeated (changes requested):
Implementation looks correct — standard RMSNorm, fp32 accumulation, eps inside the sqrt, plain weight scaling, clean shape guard. Registry wiring matches the existing attn pattern. Happy to approve once request changes are addressed.

maxiaosong1124:
> 

@KJLdefeated You're right — _manual_rms_norm uses the same expression and float order as the op, so torch.equal only proves they're copies of each other, not that the formula is correct; a shared bug (eps placement, wrong reduction dim) would pass green on both sides. Addressed in 64f6f56:

- Added test_forward_fp32_matches_torch_reference as the primary correctness check, comparing forward_fp32 against PyTorch's own F.rms_norm. Used torch.testing.assert_close(rtol=1e-6, atol=1e-6) rather than torch.equal exactly because F.rms_norm is free to reduce in a different float order — that non-identical reduction is what makes it a real independent check.
- Kept the hand-written _manual_rms_norm test as a secondary bitwise sanity check that pins the exact reference semantics.

Thanks for the careful catch!

maxiaosong1124:
> This is a solid, production-ready PR that strictly adheres to the #108 numerical contract. Here are a few professional suggestions for refinement before merging:

@Flink-ddd Good point — forward batch-invariance isn't enough; WS1 (and #153) needs the backward to be batch-invariant too, and test_gradient_flows only checked isfinite. Addressed in 64f6f56:

Added test_backward_batch_invariance_slice following your example — it runs the full-batch forward+backward, then a batch-of-1 recompute fed the matching slice of the upstream gradient, and asserts torch.equal(x_slice.grad, grad_x_full_sliced) so the input gradient is bitwise identical regardless of batch size.

On your other note (var = x_f.pow(2).mean(...) materializing a full fp32 tensor): left as-is for this PR, since you flagged it as no-change-required — readability over VRAM for the reference op. I've noted it as a watch-item for the downstream Triton/CUDA kernels on long-context workloads.

All 19 tests in tests/test_rms_norm.py pass and black --check is clean. Thanks for the review!

Flink-ddd (approved):

Flink-ddd:
the GPU CI error is not releate this PR, it's out of GPU resource. but please resolve CI-Pipeline error. also cc @KJLdefeated PTAL again.

KJLdefeated (approved):
LGTM. Approved.

maxiaosong1124:
> the GPU CI error is not releate this PR, it's out of GPU resource. but please resolve CI-Pipeline error. also cc @KJLdefeated PTAL again.

@Flink-ddd Resolved !!

Flink-ddd:
Thank you, merged.

PR #166, opened 2026-06-21
 feat(ws1): NativeSiLUOp + NativeSwiGLUOp pure-PyTorch ground-truth references + numerical contract tests

maxiaosong1124:
## Summary

Adds the pure-PyTorch ground-truth reference ops for the **gated MLP activation**
of the WS1 batch-invariant forward chain: **SiLU** (Swish) and **SwiGLU**, built on
top of the numerical contract defined in #108. Ships the two ops, their registry
wiring, docs, and a 16-case test suite that pins down both alignment axes
(Axis-A bitwise batch invariance, Axis-B per-dtype path).

Refs #108

## Terminology

This PR uses the WS1 alignment vocabulary from #108:

- **Axis-A — batch invariance (reproducibility).** A row's output must not depend on
  how many rows share the batch (batch size, slicing, padding). Asserted **bitwise**
  (`torch.equal`). This is what keeps train-time (large batch) and sample-time
  (small batch / dynamic padding) numerics identical so the policy ratio doesn't drift.
- **Axis-B — accuracy.** The low-precision (bf16 / fp16) forward path. These activations
  compute in fp32 and cast the result back to the input dtype, so the dtype output is
  **bitwise** equal to the fp32 formula cast down — asserted with `torch.equal` against
  an independent fp32 reference (no tolerance window needed for an element-wise op).

## Motivation / Context

#108 establishes the ground-truth harness and numerical contract for the WS1
batch-invariant forward chain. The Qwen3-8B dense MLP is a gated (SwiGLU) MLP:

down_proj( silu(gate_proj(x)) * up_proj(x) )

This PR covers the activation stage in the middle:

- **`silu`** — element-wise `x * sigmoid(x)` (`hidden_act="silu"`), shape-agnostic.
- **`swiglu`** — `silu(gate) * up`, where `gate` / `up` are the gate_proj / up_proj
  outputs at the intermediate dim (Qwen3-8B: **12288**). The trailing `down_proj` is a
  plain matmul and lives in a separate op.

This PR provides the deterministic fp32 reference path those downstream kernels
(Triton / CUDA / ROCm fused activation) will be validated against.

## Changes

- **`rl_engine/kernels/ops/pytorch/activation/swiglu.py`** — `NativeSiLUOp`, `NativeSwiGLUOp`
  - `forward()` — accumulate in fp32, cast result back to input dtype (Axis-B path)
  - `forward_fp32()` — fp32 accumulation, forced fp32 output (ground-truth / backward golden source)
[...]

PR #169, opened 2026-06-22
feat(ws1): NativeEmbeddingOp pure-PyTorch ground-truth reference + numerical contract tests

maxiaosong1124:
---
## Summary

Adds the pure-PyTorch ground-truth reference op for the **token embedding** — the
input layer of the WS1 batch-invariant forward chain — built on top of the numerical
contract defined in #108. Ships the op, its registry wiring, docs, and an 11-case test
suite that pins down both alignment axes (Axis-A bitwise batch invariance, Axis-B
per-dtype path), plus a GPU-only smoke test at the real Qwen3-8B table dims.

Refs #108

## Terminology

This PR uses the WS1 alignment vocabulary from #108:

- **Axis-A — batch invariance (reproducibility).** A token's output row must not depend
  on how many tokens share the batch (batch size, slicing, padding). Asserted **bitwise**
  (`torch.equal`). This is what keeps train-time (large batch) and sample-time
  (small batch / dynamic padding) numerics identical so the policy ratio doesn't drift.
- **Axis-B — accuracy.** The low-precision (bf16 / fp16) forward path. Embedding is a
  **lossless row gather** — no reduction, no fp32 accumulation — so the dtype output is
  **bitwise** equal to direct indexing at every dtype. No tolerance window is needed.

## Motivation / Context

#108 establishes the ground-truth harness and numerical contract for the WS1
batch-invariant forward chain. The first stage of the Qwen3-8B stack maps integer token
ids to their hidden-state rows:

    hidden = embedding_table[token_ids]

This PR provides the deterministic fp32 reference path that downstream kernels
(Triton / CUDA / ROCm) will be validated against. For Qwen3-8B the table is the input
embedding `[vocab=151936, hidden=4096]` and is **independent** from the lm_head weight
(`tie_word_embeddings=false`) — the two weights are not shared.

## Changes

- **`rl_engine/kernels/ops/pytorch/linear/embedding.py`** — `NativeEmbeddingOp`
  - `forward()` —  native-dtype gather, cast the gathered rows back to the weight dtype (Axis-B path)
  - `forward_fp32()` — native-dtype gather, upcast the result to fp32 (ground-truth / backward golden source)
  - Formula: out = weight[token_ids] (via F.embedding(token_ids.long(), weight))
  - Pure function — inputs never mutated in place; output dtype follows `weight`
- **`rl_engine/kernels/registry.py`** — register `PYTORCH_NATIVE_EMBEDDING` in `OpBackend`
  and add `embedding` dispatch to the cuda / rocm / cpu priority maps
- **`tests/test_embedding.py`** — 11 tests (details below)
- **`docs/operators/embedding.md`** + nav / index wiring

|---|---|
| Deterministic reference path | `forward_fp32()` gathers in fp32; tests use fixed-seed `torch.Generator` so outputs are reproducible |
| Per-dtype tolerance policy (bitwise vs tight-tolerance) | **Axis-A** asserted **bitwise** (`torch.equal`); **Axis-B** dtype output is a lossless gather, asserted **bitwise** against direct indexing (a gather needs no tolerance window) |
| Batch-config sweep / validation helper | Batch-invariance checks compute on the full batch, then assert sliced/padded rows are **bitwise identical** to their full-batch counterparts |
| Realistic shapes covered | GPU-only smoke test at the real Qwen3-8B table dims (`vocab=151936, hidden=4096`), exercising boundary ids `0` and `vocab-1`; skips when CUDA / GPU memory is unavailable |

## Test Environment

| | |
|---|---|
| OS | Ubuntu (kernel 5.15.0-124-generic) |
| Python | 3.12.3 |
| PyTorch | 2.8.0+cu128 |
| CUDA / cuDNN | 12.8 / 9.10.02 (driver 580.76.05) |
| GPU | NVIDIA vGPU-32GB |

## Test Results

```
python -m pytest tests/test_embedding.py -v
```

cf77909025a5752054fb23be11a2ff4e

The 11 tests cover:

- correctness vs direct indexing across fp32 / bf16 / fp16, asserted **bitwise**
- output shape (`token_ids.shape + (hidden,)`)
- non-int64 id tolerance (cast via `.long()`)
- Axis-A batch invariance — slice + padding variants, asserted bitwise
- purity (neither `token_ids` nor `weight` mutated in place)
- gradient flow to `weight` (fp32 autograd = backward golden source), including
  sparse-grad: rows never indexed stay exactly zero
- registry dispatch resolves `embedding` → `NativeEmbeddingOp`
- GPU-only real-shape smoke test (Qwen3-8B `vocab=151936, hidden=4096`, boundary ids)

## Checklist

- [x] Pure-PyTorch reference, no custom extension required
- [x] Covered at the real Qwen3-8B table dims (vocab=151936, hidden=4096)
- [x] Axis-A bitwise batch invariance enforced
- [x] Axis-B lossless-gather dtype path tested (bitwise, no tolerance window)
- [x] Registered in OpBackend + cuda/rocm/cpu priority maps
- [x] All 11 tests pass locally

---

## Release Notes

* **New Features**
  * Added a **Token Embedding** operator that gathers embedding vectors from integer token IDs with CPU, CUDA, and ROCm support via a native PyTorch reference backend.
  * Defined consistent **fp32 vs native-dtype** reference behavior and output dtype/shape rules.

* **Documentation**
  * Documented the operator contract (tensor shapes/dtypes, purity, and forward vs fp32 expectations), including GPU backward reproducibility limitations.

* **Tests**
  * Added extensive validation for correctness, dtype casting, non-mutation, batch/padding invariance, and gradient behavior, with CUDA-only smoke coverage for real model dimensions.


rl_engine/kernels/ops/pytorch/linear/embedding.py
```diff
@@ -0,0 +1,48 @@
...
+    # Helpers
+    # ------------------------------------------------------------------ #
+    @staticmethod
+    def _embedding(
+        token_ids: torch.Tensor,
+        weight: torch.Tensor,
+        *,
+        output_dtype: torch.dtype,
+    ) -> torch.Tensor:
+        out = F.embedding(token_ids.long(), weight.float())
```
inaniloquentee: I’m worried about this weight.float() path: it upcasts the entire embedding table before gathering, so real fp16/bf16 Qwen3-size weights will allocate an extra multi-GB fp32 copy for a tiny lookup. Since this is the only registered embedding backend today, this could make the default fallback OOM in normal GPU use.

inaniloquentee (changes requested):

rl_engine/kernels/ops/pytorch/linear/embedding.py
maxiaosong1124: @inaniloquentee  Thanks for the advice!
 The gather now runs in the weight's native dtype and only the gathered rows are upcast (F.embedding(token_ids.long(), weight).to(output_dtype)), so there's no longer a multi-GB fp32 copy of the full vocab table for a tiny lookup. Since a row gather is lossless (pure indexing, no arithmetic), this is bitwise-identical to the previous path — all 11 tests in tests/test_embedding.py still pass.

tests/test_embedding.py
```diff
@@ -0,0 +1,179 @@
...
+    op = NativeEmbeddingOp()
+    token_ids = _rand_ids((2, 4), seed=7, vocab=10)  # small vocab -> some unused rows
+    weight = _rand_weight(vocab=10, seed=7).requires_grad_(True)
+    op.forward_fp32(token_ids, weight).sum().backward()
+
+    assert torch.isfinite(weight.grad).all()
+    used = torch.unique(token_ids).tolist()
+    unused = torch.tensor([i for i in range(10) if i not in used])
+    if len(unused):
+        assert torch.equal(weight.grad[unused], torch.zeros_like(weight.grad[unused]))
```
KJLdefeated: The reduction in the backward pass: ∂L/∂weight is a scatter-add, and every repeated token id (padding, common tokens, anything) accumulates into the same weight.grad row. On CUDA, F.embedding backward does that accumulation with atomic adds, and atomic-add order is nondeterministic, so the weight gradient is not bitwise reproducible on GPU whenever ids collide. PyTorch documents embedding backward as a nondeterministic CUDA op for exactly this reason.

Either set it in the op's golden path / a fixture, or state explicitly in embedding.md under Known Limitations that the GPU backward is bitwise-reproducible only with deterministic algorithms enabled.
```python
# forward_fp32 is the backward golden source — its gradient must be
# reproducible. F.embedding backward is an atomic scatter-add on CUDA
# (nondeterministic under id collisions) unless this is set.
torch.use_deterministic_algorithms(True)
```

KJLdefeated (commented):
Overall good to me. After fix the requests, I am happy to approve.

maxiaosong1124:
> Overall good to me. After fix the requests, I am happy to approve.
@KJLdefeated Good catch — you're right that forward_fp32 is the backward golden source, so its gradient has to be reproducible, and embedding backward is a nondeterministic atomic scatter-add on CUDA whenever token ids collide (padding / common tokens accumulate into the same weight.grad row). Addressed in fab4da6 with both parts you suggested:

1. Test guard — test_embedding_gradient_flows_to_weight now wraps the backward in torch.use_deterministic_algorithms(True) (restored via try/finally) and asserts the weight gradient is bitwise identical across two independent backward passes — so the golden-source reproducibility claim is actually enforced, not just isfinite. I kept the flag in the test/fixture rather than the op's forward, since toggling that global is a process-wide side effect a reference op shouldn't own.
2. Docs — added a Known Limitations bullet in embedding.md spelling out the scatter-add / CUDA atomic-add nondeterminism and that reproducible GPU gradients require torch.use_deterministic_algorithms(True), with CPU backward always deterministic.

All 11 tests in tests/test_embedding.py still pass. Thanks for the careful review!

KJLdefeated:
LGTM.

KJLdefeated (approved):
Overall good to me. After fix the requests, I am happy to approve.

Flink-ddd:
Please resolve CI error and the code conflicts first. Thanks.

rl_engine/kernels/ops/pytorch/linear/embedding.py
```diff
@@ -0,0 +1,54 @@
...
+    """
+    Pure PyTorch native token-embedding reference.
+    out = weight[token_ids]   (a plain row gather, no accumulation)
+
+    Maps integer token ids to their hidden-state rows. For Qwen3-8B the
+    weight is the input embedding table ``[vocab=151936, hidden=4096]`` and
+    is *independent* from the lm_head weight (``tie_word_embeddings=false``).
+    """
+
+    def __init__(self) -> None:
```
Flink-ddd, line 20: same, Add torch.nn.Module to the class inheritance and initialize super().__init__().
Delete the def __call__(self, ...):

Flink-ddd (changes requested):
This PR elegantly resolves both the OOM hazard (by deferring the upcast) and the non-deterministic scatter-add edge case on CUDA. The implementation and the test coverage are rock solid.

Since the major functional edge cases have been cleared by the team, there is only one architectural standardization left to align this with the rest of the WS1 reference suite.

rl_engine/kernels/ops/pytorch/linear/embedding.py
maxiaosong1124: > same, Add torch.nn.Module to the class inheritance and initialize super().**init**(). Delete the def **call**(self, ...):

 @Flink-ddd Addressed in d8b3d39 — NativeEmbeddingOp now inherits nn.Module, __init__ calls super().__init__(), and the explicit __call__ is removed (handled by nn.Module). All tests pass and CI is green locally. Also merged latest main (resolved the registry conflict from #166).

Flink-ddd (approved):
LGTM, Thank you for update, Let's merge this PR.

Flink-ddd:
cc @inaniloquentee PTAL, if there are no further issue, we will merge it.

rl_engine/kernels/ops/pytorch/linear/embedding.py
inaniloquentee: > @inaniloquentee Thanks for the advice! The gather now runs in the weight's native dtype and only the gathered rows are upcast (F.embedding(token_ids.long(), weight).to(output_dtype)), so there's no longer a multi-GB fp32 copy of the full vocab table for a tiny lookup. Since a row gather is lossless (pure indexing, no arithmetic), this is bitwise-identical to the previous path — all 11 tests in tests/test_embedding.py still pass.

LGTM now.

inaniloquentee (approved):

PR #170, opened 2026-06-22
feat(ws1): NativeLMHeadOp pure-PyTorch ground-truth reference + numerical contract tests

maxiaosong1124:
## Summary

Adds the pure-PyTorch ground-truth reference op for the **language-model head** — the
output layer of the WS1 batch-invariant forward chain — built on top of the numerical
contract defined in [#108](https://github.com/RL-Align/RL-Kernel/issues/108). Ships the
op, its registry wiring, docs, and a 15-case test suite that pins down both alignment
axes (Axis-A bitwise batch invariance, Axis-B per-dtype accuracy), plus a GPU-only smoke
test at the real Qwen3-8B projection dims.

Refs [#108](https://github.com/RL-Align/RL-Kernel/issues/108)

## Terminology

This PR uses the WS1 alignment vocabulary from [#108](https://github.com/RL-Align/RL-Kernel/issues/108):

- **Axis-A — batch invariance (reproducibility).** A row's logits must not depend on how
  many rows share the batch (batch size, slicing, padding). Asserted **bitwise**
  (`torch.equal`). This is what keeps train-time (large batch) and sample-time (small
  batch / dynamic padding) numerics identical so the policy ratio doesn't drift.
- **Axis-B — accuracy.** The low-precision (bf16 / fp16) forward path. Unlike the
  lossless embedding gather, lm_head is a **reduction over hidden**, so low-precision
  accumulation drifts from the fp32 reference and is checked against a **tolerance**
  window (not bitwise).

## Motivation / Context

[#108](https://github.com/RL-Align/RL-Kernel/issues/108) establishes the ground-truth
harness and numerical contract for the WS1 batch-invariant forward chain. The final stage
of the Qwen3-8B stack projects hidden states back to vocabulary logits:

logits = hidden @ weight.t()   # weight is HF [out, in]

This PR provides the deterministic fp32 reference path that downstream kernels (Triton /
CUDA / ROCm) will be validated against. For Qwen3-8B the weight is the output projection
`[vocab=151936, hidden=4096]` in the HF `nn.Linear` `[out, in]` convention (transposed
internally), is **independent** from the embedding table (`tie_word_embeddings=false`),
and has **no bias**.

## Changes

- **`rl_engine/kernels/ops/pytorch/linear/lm_head.py`** — `NativeLMHeadOp`
  - `forward()` — project in the input dtype, output the input dtype (Axis-B path)
  - `forward_fp32()` — upcast to fp32, accumulate in fp32, forced fp32 output
    (ground-truth / backward golden source)
  - Formula: `out = hidden @ weight.t() (+ bias)`
  - Weight is HF `[out, in]` and transposed internally — the one difference from the bare
    `matmul` op; do not use interchangeably
  - Pure function — inputs never mutated in place; output dtype follows `hidden`
- **`rl_engine/kernels/registry.py`** — register `PYTORCH_NATIVE_LM_HEAD` in `OpBackend`
  and add `lm_head` dispatch to the cuda / rocm / cpu priority maps
- **`tests/test_lm_head.py`** — 15 tests (details below)
- **`docs/operators/lm_head.md`** + nav / index wiring

|---|---|
| Deterministic reference path | `forward_fp32()` accumulates in fp32; tests use fixed-seed `torch.Generator` so outputs are reproducible |
| Per-dtype tolerance policy (bitwise vs tight-tolerance) | **Axis-A** asserted **bitwise** (`torch.equal`); **Axis-B** reduction drift checked against a per-dtype tolerance measured relative to the output peak (bf16 ~0.37%, fp16 ~0.05%) |
| Batch-config sweep / validation helper | Batch-invariance checks compute on the full batch, then assert sliced/padded rows are **bitwise identical** to their full-batch counterparts, across fp32 / bf16 / fp16 |
| Realistic shapes covered | GPU-only smoke test at the real Qwen3-8B dims (`vocab=151936, hidden=4096`); CPU tests keep the real `hidden=4096` reduction length (only vocab is shrunk) so the drift is representative; skips when CUDA / GPU memory is unavailable |

## Reduction-specific note (Axis-A reduction order)

A single `torch.matmul` is **not** bitwise batch-invariant by default: multi-threaded CPU
GEMM splits the `hidden` (K) reduction across threads by the `M = batch*seq` dimension, so
"compute full then slice" ≠ "compute slice" once `hidden` is large. The tests pin a single
thread to fix the reduction order (a local stand-in for the planned
`testing/determinism.py::deterministic_context`). On GPU cuBLAS likewise splits K by `M`,
so a batch-invariant GEMM is a downstream kernel concern — the GPU smoke test validates the
full-vocab shape and fp32 correctness, not Axis-A bitwise.

## Test Environment

| | |
|---|---|
| OS | Ubuntu (kernel 5.15.0-124-generic) |
| Python | 3.12.3 |
| PyTorch | 2.8.0+cu128 |
| CUDA / cuDNN | 12.8 / 9.10.02 |

## Test Results

python -m pytest tests/test_lm_head.py -v
17 passed
image

The 17 tests cover:

- fp32 correctness vs naive matmul, asserted **bitwise** (`torch.equal`); fp32 forward
  path bitwise-equal to the ground truth
- bf16 / fp16 dtype-path accuracy — max abs error relative to output peak (bf16 ~0.37%,
  fp16 ~0.05%), with error stats printed
- output shape (`hidden.shape[:-1] + (vocab,)`)
- bias semantics (`None` default == no bias; provided `[vocab]` bias added)
- Axis-A batch invariance — slice + padding variants across fp32 / bf16 / fp16, asserted
  bitwise under a pinned single-thread reduction
- purity (neither `hidden`, `weight`, nor `bias` mutated in place)
- gradient flow to `hidden` / `weight` (fp32 autograd = backward golden source), verified
  against the closed-form grads
- registry dispatch resolves `lm_head` → `NativeLMHeadOp`
- GPU-only real-shape smoke test (Qwen3-8B `vocab=151936, hidden=4096`)

## Checklist

- ✅ Pure-PyTorch reference, no custom extension required
- ✅ Covered at the real Qwen3-8B projection dims (vocab=151936, hidden=4096)
- ✅ Axis-A bitwise batch invariance enforced (fixed single-thread reduction order)
- ✅ Axis-B per-dtype tolerance calibrated (relative-to-peak, stats in PR)
- ✅ Registered in OpBackend + cuda/rocm/cpu priority maps
- ✅ All 17 tests pass locally

## Summary by CodeRabbit

* **New Features**
  * Added documentation for the LM Head operator, including usage, supported shapes/dtypes, and precision behavior.
  * Registered LM Head as a supported operator so it can be selected automatically.
* **Bug Fixes**
  * Improved numeric consistency for LM Head inference and reference outputs across precision modes.
* **Tests**
  * Added coverage for output shape, bias handling, gradients, batch/padding invariance, and backend dispatch.


rl_engine/kernels/ops/pytorch/linear/lm_head.py
```diff
@@ -0,0 +1,93 @@
...
+    def _lm_head(
+        hidden: torch.Tensor,
+        weight: torch.Tensor,
+        bias: Optional[torch.Tensor],
+        *,
+        compute_dtype: torch.dtype,
+        output_dtype: torch.dtype,
+    ) -> torch.Tensor:
+        """Core matmul: cast to ``compute_dtype``, project, optionally add bias, cast out."""
+        h = hidden.to(compute_dtype)
```
inaniloquentee, line 102: One precision nit: forward_fp32() casts the inputs to fp32, but the matmul can still run under autocast or CUDA TF32 settings, so it may not be a true fp32 golden reference. Since downstream kernels will compare against this path, could we explicitly disable autocast/TF32 around the matmul, or document that required precision context?

inaniloquentee (changes requested):

rl_engine/kernels/ops/pytorch/linear/lm_head.py
maxiaosong1124: Thanks, @inaniloquentee .

  forward_fp32() now disables autocast and CUDA TF32 around the matmul, while saving/restoring the previous TF32 setting so global state does not leak. The regular forward() path is unchanged and still
  follows the ambient precision context.

  I also added regression coverage for this:
  - CPU autocast case: forward_fp32 remains equal to the fp32 reference and restores the TF32 flag.
  - CUDA TF32 case: forward_fp32 is checked against a higher-precision reference when CUDA is available.

  Docs were updated to note the precision-context behavior.

tests/test_lm_head.py
```diff
@@ -0,0 +1,327 @@
...
+    hidden = _rand_hidden(2, 4, seed=8).requires_grad_(True)
+    weight = _rand_weight(seed=8).requires_grad_(True)
+    op.forward_fp32(hidden, weight).sum().backward()
+
+    assert torch.isfinite(hidden.grad).all() and torch.isfinite(weight.grad).all()
+    assert hidden.grad.shape == hidden.shape and weight.grad.shape == weight.shape
+    exp_h = weight.detach().sum(dim=0).expand_as(hidden.grad)
+    exp_w = hidden.detach().sum(dim=(0, 1)).expand_as(weight.grad)
+    assert torch.allclose(hidden.grad, exp_h, atol=1e-4, rtol=1e-4)
+    assert torch.allclose(weight.grad, exp_w, atol=1e-4, rtol=1e-4)
```
KJLdefeated: Closed-form check is correct and a real improvement over finiteness, but the out.sum() cotangent is the least discriminating one possible: it makes the upstream gradient all-ones, so both expected grads collapse to column sums (every row of hidden.grad identical, every row of weight.grad identical). A backward that transposed or mis-contracted something can still pass under that symmetry, and the real backward never sees an all-ones cotangent.
Recommend backpropping a random cotangent so the full contraction is exercised:
```python
dy = torch.randn_like(out)
out.backward(dy)
exp_h = dy @ weight                                   # [.., V] @ [V, K] -> [.., K]
exp_w = dy.reshape(-1, V).t() @ hidden.reshape(-1, K) # [V, K]
torch.testing.assert_close(hidden.grad, exp_h, rtol=1e-5, atol=1e-5)
torch.testing.assert_close(weight.grad, exp_w, rtol=1e-5, atol=1e-5)
```

KJLdefeated (changes requested):
Overall looks good and clean. I am happy to approve once the requests are resolved.

tests/test_lm_head.py
maxiaosong1124: @KJLdefeated Fixed in 407200a — switched to a random (fixed-seed) cotangent and assert against the exact closed forms (dy @ weight / dy^T @ hidden), so the full contraction is exercised instead of collapsing to column sums. Seeded for reproducibility; all 17 tests pass at 1e-5. Also rebased on latest main and got CI linting back to green.

rl_engine/kernels/ops/pytorch/linear/lm_head.py
inaniloquentee: > Thanks, @inaniloquentee .
> 
> forward_fp32() now disables autocast and CUDA TF32 around the matmul, while saving/restoring the previous TF32 setting so global state does not leak. The regular forward() path is unchanged and still follows the ambient precision context.
> 
> I also added regression coverage for this:
> 
> * CPU autocast case: forward_fp32 remains equal to the fp32 reference and restores the TF32 flag.
> * CUDA TF32 case: forward_fp32 is checked against a higher-precision reference when CUDA is available.
> 
> Docs were updated to note the precision-context behavior.

LGTM!

inaniloquentee (approved):

Flink-ddd (approved):
LGTM, cc @KJLdefeated if there are no further issues, please approve this PR and we will merge it.

Flink-ddd:
@maxiaosong1124 please resolve CI error first.

maxiaosong1124:
> @maxiaosong1124 please resolve CI error first.

@Flink-ddd CI error has been resolved!

KJLdefeated (approved):
Approved

Issue #112, opened 2026-06-15
[WS2] Deterministic NCCL all-reduce (incl. DP gradient all-reduce)


Flink-ddd:
Part of #83

Make NCCL all-reduce a deterministic operation across the WS2 ops, including the DP-path gradient all-reduce. Independent of WS1 ops — can start in W1.

Planned PRs:
- [ ] Audit and enumerate all all-reduce call sites (forward and backward).
- [ ] NVLink-Sharp in-switch deterministic reduce on Hopper (CUDA 12.8+) where available.
- [ ] Fallback for non-NVSwitch setups: specify the concrete path (e.g. forced single ring all-reduce with fixed rank order, or deterministic reduce-scatter + all-gather) — not just "a fallback."
- [ ] DP gradient all-reduce determinism: make the cross-DP-rank gradient all-reduce order deterministic; verify gradients match across DP=1 vs DP>1 on a fixed step (a distinct, training-side drift scenario).
- [ ] Doc.

Can start in W1 (does not depend on WS1 ops).

CyberSecurityErial:
Hi @inaniloquentee, I’m interested in helping with the NCCL / communication determinism part of #112.

I can start with a small audit/design PR:
1. enumerate all collective call sites involved in WS2;
2. classify them by forward/backward path and TP/SP/DP/FSDP group;
3. document which reductions can affect logprob/loss/KL/gradient drift;
4. propose deterministic fallback paths such as fixed-rank single-ring and deterministic reduce-scatter + all-gather;
5. add NCCL/topology/rank-order debug metadata recommendations.

Would this overlap with your current plan, or is this a useful slice to take?

inaniloquentee:
> Hi [@inaniloquentee](https://github.com/inaniloquentee), I’m interested in helping with the NCCL / communication determinism part of [#112](https://github.com/RL-Align/RL-Kernel/issues/112).
> 
> I can start with a small audit/design PR:
> 
> 1. enumerate all collective call sites involved in WS2;
> 2. classify them by forward/backward path and TP/SP/DP/FSDP group;
> 3. document which reductions can affect logprob/loss/KL/gradient drift;
> 4. propose deterministic fallback paths such as fixed-rank single-ring and deterministic reduce-scatter + all-gather;
> 5. add NCCL/topology/rank-order debug metadata recommendations.
> 
> Would this overlap with your current plan, or is this a useful slice to take?

[...]

PR #177, opened 2026-06-22
[readme]update Architecture image

Flink-ddd:
update Architecture image, Add support for specific domestically produced chips. 

The CI errors caused by .github/workflows/auto-reviewer.yml file have been deleted because we already have the .github/CODEOWNERS mechanism.

PR #181, opened 2026-06-23
[WS2][distributed] Add deterministic all-reduce contract and smoke tests

CyberSecurityErial:
## Summary

Part of #112.

This PR adds the first deterministic all-reduce slice for WS2:

- audit current all-reduce / collective call sites;
- add a small `rl_engine.distributed` all-reduce helper;
- add an ordered-rank reference mode for tests and debugging;
- add CPU/Gloo smoke tests;
- add a fixed-batch DP gradient comparison smoke test;
- document the supported modes and current limitations.

## API

`deterministic_all_reduce(tensor, config)` reduces the input tensor in place.

Modes:

- `torch_all_reduce`: uses `torch.distributed.all_reduce`.
- `ordered_rank_reference`: gathers rank-local tensors, accumulates them on process-group rank 0 in process-group rank order, then broadcasts the result.

The NCCL environment helper is best-effort and must run before process-group initialization. It does not claim NVLS or hardware-wide bitwise determinism.

## Validation

```bash
ruff check rl_engine/distributed tests/distributed
python -m mypy --ignore-missing-imports rl_engine/distributed
PYTEST_DISABLE_PLUGIN_AUTOLOAD=1 python -m pytest tests/distributed/test_deterministic_allreduce.py -q
PYTEST_DISABLE_PLUGIN_AUTOLOAD=1 python -m pytest tests/distributed/test_dp_gradient_determinism.py -q
black --check --line-length 100 rl_engine/distributed tests/distributed
isort --check-only --profile black --line-length 100 rl_engine/distributed tests/distributed
flake8 --max-line-length=100 --extend-ignore=E203,E704 rl_engine/distributed tests/distributed
mkdocs build --strict -f mkdocs.yaml
git diff --check
```

Results:

```text
all-reduce pytest smoke: 5 passed
DP gradient pytest smoke: 2 passed
```

Manual CUDA/NCCL smoke on the local 2-rank CUDA 12.8 environment:

```bash
CUDA_VISIBLE_DEVICES=0,1 NCCL_ALGO=Ring NCCL_PROTO=Simple NCCL_MIN_NCHANNELS=1 NCCL_MAX_NCHANNELS=1 \
torchrun --standalone --nproc_per_node=2 \
  tests/distributed/test_deterministic_allreduce.py \
  --backend nccl --mode torch_all_reduce --dtype fp32 --device cuda --iterations 3

CUDA_VISIBLE_DEVICES=0,1 NCCL_ALGO=Ring NCCL_PROTO=Simple NCCL_MIN_NCHANNELS=1 NCCL_MAX_NCHANNELS=1 \
torchrun --standalone --nproc_per_node=2 \
  tests/distributed/test_dp_gradient_determinism.py \
  --backend nccl --mode torch_all_reduce --dtype fp32 --device cuda
```

Results:

```text
2-rank CUDA/NCCL all-reduce smoke: pass, bitwise equal
2-rank CUDA/NCCL DP gradient smoke: pass, max_abs_diff=5.960464477539063e-08
```

## Limitations

- NVLS / NVLink-Sharp is not implemented or claimed in this PR.
- Multi-node and RDMA behavior were not validated.
- DeepSpeed internal gradient communication order is not controlled yet.

## Summary by CodeRabbit

* **New Features**
  * Added deterministic distributed all-reduce support for tensors, including configurable reduction modes and optional NCCL determinism setup.
* **Documentation**
  * Added developer guide pages covering deterministic all-reduce behavior and an audit of where distributed collectives may be involved.
  * Updated documentation navigation to include the new “Distributed” subsection.
* **Tests**
  * Added smoke tests and determinism checks for deterministic all-reduce.
  * Added gradient determinism validation comparing baseline vs partitioned DP-style training runs.


CyberSecurityErial:
Reviews may only be requested from collaborators.

  The workflow uses a hardcoded reviewer list, and at least one listed user is not a collaborator of RL-Align/RL-Kernel. This happens before any code/test validation from this PR.

  I noticed #177 removes .github/workflows/auto-reviewer.yml, which should address this repository-level issue. Could maintainers ignore/rerun this check after the workflow fix, or mark it non-blocking for this PR?

rl_engine/distributed/deterministic_allreduce.py
```diff
@@ -0,0 +1,147 @@
...
+    if cfg.op == "mean" and not (tensor.is_floating_point() or tensor.is_complex()):
+        raise TypeError("op='mean' requires a floating-point or complex tensor")
+
+
+def _torch_all_reduce(
+    tensor: torch.Tensor,
+    cfg: DeterministicAllReduceConfig,
+    world_size: int,
+) -> torch.Tensor:
+    dist.all_reduce(tensor, op=dist.ReduceOp.SUM, group=cfg.group, async_op=False)
```
inaniloquentee, line 101: torch_all_reduce can silently give wrong results for non-contiguous tensors. I reproduced it with a 2-rank Gloo run on a base[:, ::2] view. Could we make it use a contiguous buffer and copy back, or reject non-contiguous inputs? A small smoke test would cover this.

inaniloquentee (changes requested):

rl_engine/distributed/deterministic_allreduce.py
CyberSecurityErial: Thanks for catching this. I added a contiguity check for `torch_all_reduce` and a 2-rank Gloo smoke test covering a `base[:, ::2]` view in a new pr.

rl_engine/distributed/deterministic_allreduce.py
```diff
@@ -0,0 +1,147 @@
...
+        reduced = gathered[0].to(dtype=dtype)
+        for item in gathered[1:]:
+            reduced.add_(item.to(dtype=dtype))
+        if cfg.op == "mean":
+            reduced.div_(world_size)
+        result.copy_(reduced.to(dtype=send.dtype))
+
+    dist.broadcast(result, src=_group_root_global_rank(cfg.group), group=cfg.group)
+    tensor.copy_(result.view_as(tensor))
+    return tensor
```
KJLdefeated, line 128: `ordered_rank_reference` genuinely deterministic (fixed rank-order fp32 accumulation), but all_gather materializes world_size copies of the tensor on every rank. For a real gradient or logit tensor that's O(world_size) memory per rank which is a small-tensor oracle, not a production reduction. The DP smoke test only runs TinyGradientModel, so this ceiling never gets exercised.

rl_engine/distributed/deterministic_allreduce.py
```diff
@@ -0,0 +1,147 @@
...
+
+def _torch_all_reduce(
+    tensor: torch.Tensor,
+    cfg: DeterministicAllReduceConfig,
+    world_size: int,
+) -> torch.Tensor:
+    dist.all_reduce(tensor, op=dist.ReduceOp.SUM, group=cfg.group, async_op=False)
+    if cfg.op == "mean":
+        tensor.div_(world_size)
+    return tensor
```
KJLdefeated, line 104: `force_fp32_accumulation` is silently ignored here. It defaults True, so the default config (torch_all_reduce + force_fp32=True) advertises fp32 accumulation but NCCL reduces in native bf16/fp16.
I would suggest this change:
```python
def _torch_all_reduce(tensor, cfg, world_size):
    if not tensor.is_contiguous():
        tensor_c = tensor.contiguous()          # NCCL/Gloo require contiguous buffers
    else:
        tensor_c = tensor
    if cfg.force_fp32_accumulation and tensor_c.is_floating_point() and tensor_c.dtype not in (torch.float32, torch.float64):
        acc = tensor_c.float()
        dist.all_reduce(acc, op=dist.ReduceOp.SUM, group=cfg.group, async_op=False)
        if cfg.op == "mean":
            acc.div_(world_size)
        tensor.copy_(acc.to(tensor.dtype))       # copy back into original (possibly non-contiguous) tensor
        return tensor
    dist.all_reduce(tensor_c, op=dist.ReduceOp.SUM, group=cfg.group, async_op=False)
    if cfg.op == "mean":
        tensor_c.div_(world_size)
    if tensor_c is not tensor:
        tensor.copy_(tensor_c)
    return tensor
```

Issue #42, opened 2026-05-31
[FEAT][kernels]: implement Fused Masking and Variable-Length Sequence Packing (Pack-and-Pad)


Flink-ddd:
### Context: 
During RL training (PPO/DPO/GRPO), only tokens in the generated response and non-padding tokens contribute to the loss. Computing and storing full-sequence logits including the prompt wastes massive VRAM. We need an operator that efficiently filters out inactive tokens and packs the active tokens into a contiguous flat 1D array before heavy loss computation.  

## Tasks:  
1. Design a CUDA/Triton kernel that takes a dense tensor [B, S, ...], an active boolean mask [B, S], and compacts it into a 1D tensor [Total_Active, ...].  
2. Implement the corresponding Unpack scatter operator to restore the gradient back to the original [B, S, ...] shape during the backward pass if needed.  
3. Optimize index computation using cumulative sum (prefix sum) algorithms (e.g., Thrust or custom Triton scan).

Chen-BUPT:
i`d like to have a try

Flink-ddd:
Sure, Assigned, thank you.

PR #182, opened 2026-06-23
feat(kernels): add fused masking + variable-length pack-and-pad op 

Chen-BUPT:
# Fused masking + variable-length pack-and-pad op (#42)

## What this adds

A `pack` operator that compacts the **active** rows of a dense `[B, S, *tail]`
tensor (selected by a `[B, S]` mask) into a contiguous `[Total_Active, *tail]`
tensor, returns per-row `cu_seqlens`, and scatters gradients back to the dense
layout on the backward pass.

| Backend | Class | Notes |
|---|---|---|
| PyTorch (CPU/GPU fallback) | `NativePackOp` | Portable reference; defines the numerical contract. Packing order matches `SyntheticRLKernelBatch.compact_completion_values`. |
| Triton (CUDA & ROCm) | `TritonPackOp` | Host-side prefix-sum for destination indices; gather/scatter kernels move the tail vectors. Numerically identical to the native op. |

Registry dispatch for `"pack"`: Triton on GPU, PyTorch fallback on CPU.

## Correctness

- 18/18 tests in `tests/test_pack.py` pass on an **NVIDIA H20 (SM90, CUDA 13.0)**.
- Native op validated against the repo's canonical compaction + `gradcheck`.
- Triton op validated against the native op (forward + backward); **max-abs
  drift = `0.000e+00`** across all benchmarked shapes.

## Why it matters: end-to-end VRAM

In RL training only the response / non-padding tokens contribute to the loss.
Packing the hidden states **before** the vocab projection means the full
`[B, S, V]` logits are never materialized for masked-out tokens — exactly the
saving #42 targets ("saved memory can be used for larger batches or longer
CoT").

Benchmark: `hidden=4096 -> lm_head -> logits -> selected logp`, bf16,
`B=32, S=1024`, comparing dense (full logits) vs pack-then-project.

### vocab = 131072

| mask_density | valid_tokens | dense logp (GB) | packed logp (GB) | **VRAM saving** | speedup | pack_drift |
|---|---|---|---|---|---|---|
| 0.05 | 1638 | 17.31 | 2.12 | **87.8 %** | 0.23 | 0 |
| 0.10 | 3277 | 17.33 | 2.96 | **82.9 %** | 0.29 | 0 |
| 0.30 | 9830 | 17.43 | 6.31 | **63.8 %** | 0.45 | 0 |
| 0.50 | 16384 | 17.53 | 9.66 | **44.9 %** | 0.57 | 0 |
| 1.00 | 32768 | 17.78 | 18.03 | -1.4 % | 0.74 | 0 |

### vocab = 32768

| mask_density | valid_tokens | dense logp (GB) | packed logp (GB) | **VRAM saving** | speedup | pack_drift |
|---|---|---|---|---|---|---|
| 0.05 | 1638 | 4.56 | 0.77 | **83.1 %** | 0.25 | 0 |
| 0.10 | 3277 | 4.58 | 1.01 | **78.0 %** | 0.32 | 0 |
| 0.30 | 9830 | 4.68 | 1.96 | **58.2 %** | 0.46 | 0 |
| 0.50 | 16384 | 4.78 | 2.91 | **39.2 %** | 0.54 | 0 |
| 1.00 | 32768 | 5.03 | 5.28 | -5.0 % | 0.78 | 0 |

(Full data: [`pack_h20.csv`](./pack_h20.csv).)

## How to read the numbers

- **VRAM saving is the headline.** With a sparse loss mask (long prompt, short
  response — the common RL case), packing before the projection cuts peak logp
  memory by up to **~88 %** (density 0.05). The `17 GB -> 2 GB` drop lets the
  same GPU fit a much larger batch or longer CoT.
- **density = 1.0 (all active) is the control group:** saving ≈ 0 % (a small
  pack overhead), as expected — there is nothing to compact, so nothing is
  saved. This confirms the measurement is honest.
- **Latency (`speedup` < 1) is reported as-is.** The pack op itself is
  memory-bound and a touch slower than PyTorch's boolean indexing
  (`index_select` is already highly tuned). Its absolute cost is 0.06–0.35 ms,
  negligible against the multi-GB memory it saves. This PR targets the VRAM
  win (#42's stated motivation); 

## Reproduce

```bash
PYTHONPATH=. python -m pytest tests/test_pack.py -v        # correctness (needs CUDA for Triton cases)

PYTHONPATH=. python benchmarks/benchmark_pack.py \
  --num-prompts 4 --g-sizes 8 --hidden-dim 4096 \
  --mask-densities 0.05,0.1,0.3,0.5,1.0 \
  --completion-lens 1024 --vocab-sizes 32768,131072 \
  --output benchmarks/results/pack_h20.csv
```

## Notes for reviewers

- Hardware fallback: no GPU / no Triton → automatically dispatches to
  `NativePackOp` (CPU path); Triton tests skip cleanly.
- ROCm: the Triton kernels are wavefront-agnostic (plain gather/scatter), so
  they run on ROCm via Triton without CUDA-specific intrinsics.
- Please add the **`needs-gpu-ci`** label so the GPU CI exercises the Triton
  path.

## Summary by CodeRabbit

* **New Features**
  * Added pack-and-pad support to compact active sequence tokens into packed tensors with sequence-length metadata.
  * Added unpack support to restore packed data to dense layouts, including multidimensional data.
  * Enabled pack operation dispatch across CPU, CUDA, and ROCm environments.
  * Added a benchmark runner with smoke-test mode and CSV output.

* **Documentation**
  * Added operator documentation, navigation links, usage details, and benchmarking instructions.

* **Tests**
  * Added coverage for masking, gradients, round trips, edge cases, validation, and registry dispatch.


rl_engine/kernels/ops/pytorch/packing/pack.py
```diff
@@ -0,0 +1,108 @@
...
+        tail_shape = x.shape[lead:]
+
+        flat_mask = mask.reshape(-1).to(torch.bool)
+        flat_x = x.reshape(-1, *tail_shape) if tail_shape else x.reshape(-1)
+
+        index = flat_mask.nonzero(as_tuple=False).squeeze(-1)
+        packed = flat_x.index_select(0, index)
+
+        # cu_seqlens: prefix-sum of per-row active counts, for varlen consumers.
+        per_row_active = mask.reshape(mask.shape[0], -1).to(torch.int64).sum(dim=1)
```
inaniloquentee: packed uses mask.to(bool), but cu_seqlens sums the raw mask. Non-bool masks can produce wrong prefix sums. Please either require bool masks or compute counts from the bool mask.

inaniloquentee (changes requested):

rl_engine/kernels/ops/triton/packing/pack.py
```diff
@@ -0,0 +1,156 @@
...
+
+
+def _dest_index(flat_mask: torch.Tensor) -> Tuple[torch.Tensor, int]:
+    """Map each flattened row to its packed destination index (active rows get an
+    exclusive prefix-sum position; inactive rows get -1). Returns (dest, n_active)."""
+    active = flat_mask.to(torch.bool)
+    counts = active.to(torch.int64)
+    # Exclusive prefix sum: position of each active row in the packed buffer.
+    excl = torch.cumsum(counts, dim=0) - counts
+    n_active = int(counts.sum().item())
```
Flink-ddd: (_dest_index, n_active = int(counts.sum().item())) .item() forces a device-to-host sync that blocks the stream, since torch.empty(n_active, ...) depends on this Python int. It's hard to avoid given the output shape depends on n_active, but please add a comment marking this as an intentional sync point so it isn't duplicated on the hot path later.

rl_engine/kernels/ops/triton/packing/pack.py
```diff
@@ -0,0 +1,156 @@
...
+        src = x.reshape(n_rows, T).contiguous()
+        flat_mask = mask.reshape(-1)
+        dest, n_active = _dest_index(flat_mask)
+
+        packed = torch.empty(n_active, T, device=x.device, dtype=x.dtype)
+        if n_active > 0:
+            grid = (n_rows, triton.cdiv(T, _BLOCK_T))
+            _pack_gather_kernel[grid](src, packed, dest, T, BLOCK_T=_BLOCK_T)
+
+        per_row_active = flat_mask.to(torch.bool).reshape(mask.shape[0], -1).to(torch.int64).sum(1)
```
Flink-ddd: (per_row_active = ...reshape(mask.shape[0], -1)...) cu_seqlens treats mask.shape[0] as the batch dim, but packing supports an arbitrary-dim leading mask. For mask.dim() > 2 the per-row grouping no longer matches the row-major pack order. Please document that cu_seqlens is only well-defined for a 2D [B, S] mask.

rl_engine/kernels/ops/pytorch/packing/pack.py
```diff
@@ -0,0 +1,111 @@
...
+            f"mask shape {tuple(mask.shape)} must match the leading dims of "
+            f"x.shape {tuple(x.shape)} (expected {tuple(x.shape[: mask.dim()])})."
+        )
+
+
+class _PackFunction(torch.autograd.Function):
+    """forward: gather active rows; backward: scatter grads to active rows."""
+
+    @staticmethod
+    def forward(ctx, x: torch.Tensor, mask: torch.Tensor):
```
Flink-ddd, line 45: (_PackFunction.forward / _validate) If x and mask are on different devices, index_select throws a low-level device-mismatch error. Please add a device check in _validate for a clear message, consistent with the shape check.

rl_engine/kernels/ops/pytorch/packing/pack.py
```diff
@@ -0,0 +1,111 @@
...
+    contiguous ``[Total_Active, *tail]`` tensor and returns the per-row
+    ``cu_seqlens`` prefix-sum. Backward scatters the upstream gradient back to
+    the original ``[B, S, *tail]`` layout, leaving zeros at inactive positions.
+    """
+
+    def __call__(self, x: torch.Tensor, mask: torch.Tensor) -> Tuple[torch.Tensor, torch.Tensor]:
+        return _PackFunction.apply(x, mask)
+
+    @staticmethod
+    def unpack(
```
Flink-ddd, line 115: (NativePackOp.unpack) When tail_shape is passed explicitly and the element count happens to match but dims differ, reshape silently produces a wrong shape. Consider asserting tail_shape against packed.shape[1:].

benchmarks/benchmark_pack.py
```diff
@@ -0,0 +1,396 @@
...
+        help="Active-token fraction; sparse masks show the largest VRAM saving",
+    )
+    parser.add_argument("--seed", type=int, default=0)
+    parser.add_argument("--warmup", type=int, default=3)
+    parser.add_argument("--repeat", type=int, default=10)
+    parser.add_argument("--output", type=Path, default=None)
+    return parser
+
+
+def main() -> None:
```
Flink-ddd, line 356: (main(), OOM handler) The OOM-row dict duplicates the structure of _pack_row's return value, so it's easy to miss when CSV_COLUMNS changes. Consider a _blank_row(config, status, notes) helper.

tests/test_pack.py
```diff
@@ -0,0 +1,274 @@
...
+def test_triton_supports_multidim_tail():
+    mask = torch.tensor([[True, False, True], [False, True, True]], device="cuda")
+    x = torch.randn(2, 3, 4, 5, device="cuda")
+    packed_t, cu_t = TritonPackOp()(x, mask)
+    packed_n, cu_n = NativePackOp()(x, mask)
+    assert packed_t.shape == (4, 4, 5)
+    assert torch.equal(packed_t, packed_n)
+    assert cu_t.tolist() == [0, 2, 4]
+
+
```
Flink-ddd: Add a negative test where x and mask are on different devices, asserting a clear error is raised (matches the native/triton device-check suggestion above).

tests/test_pack.py
```diff
@@ -0,0 +1,274 @@
...
+@requires_triton_cuda
+@pytest.mark.parametrize("valid_density", [1.0, 0.7, 0.0])
+def test_triton_forward_matches_native(valid_density):
+    batch = _batch(seed=10, device="cuda", valid_density=valid_density)
+    x = _dense(batch, seed=110, device="cuda")
+    packed_t, cu_t = TritonPackOp()(x, batch.completion_mask)
+    packed_n, cu_n = NativePackOp()(x, batch.completion_mask)
+    assert torch.equal(packed_t, packed_n)
+    assert torch.equal(cu_t, cu_n)
+
```
Flink-ddd: (Triton tests) Add a dedicated n_active == 0 test asserting the Triton path doesn't launch the kernel (the if n_active > 0 guard) and returns the correct [0, ...] shape.

Flink-ddd (approved):
Excellent work! but need add related doc file and Please revise the following review comments. Everything else is very well done. Once these are done, I will approve it. Thanks.

Flink-ddd (changes requested):
Correction: my previous review was approved by mis-click. The doc and the review comments above still need to be addressed before this can be merged re-submitting as Request changes. Will approve once they're done. Thanks!

rl_engine/kernels/ops/triton/packing/pack.py
```diff
@@ -0,0 +1,156 @@
...
+
+class TritonPackOp:
+    """Triton fused masking + variable-length packing (pack-and-pad).
+
+    Forward packs the active rows of ``x`` (selected by ``mask``) into a
+    contiguous ``[Total_Active, *tail]`` tensor and returns the per-row
+    ``cu_seqlens`` prefix-sum. Backward scatters the upstream gradient back to the
+    original ``[*mask.shape, *tail]`` layout, leaving zeros at inactive positions.
+    Numerically identical to ``NativePackOp``; CUDA & ROCm via Triton.
+    """
```
KJLdefeated: What does the Triton path buy over index_select today? Reading the benchmark honestly: TritonPackOp runs 1.3–4× slower than the PyTorch boolean index baseline (speedup 0.23–0.78), and its output and peak memory are identical to NativePackOp. It's the same copy. So as it stands the Triton kernel is strictly dominated by x[mask] / the native op on every axis.
The only justification I can see is future fusion, folding this gather into the downstream lm_head/logp kernel so the packed hidden is never materialized as a standalone tensor. But that fused consumer isn't in this PR, so right now we'd be committing to maintain a slower-than-native CUDA/ROCm kernel with no caller that needs it.

KJLdefeated (changes requested):
The native op is clean and correct gather/scatter, race-free (unique dest indices, no atomics), drift-free by construction. Aligning with Flink's existing requested-changes (device check, unpack tail_shape assert, cu_seqlens-is-2D-only doc, _blank_row helper, the negative-device test — all fair, should land).

Flink-ddd (approved):

Flink-ddd:
@KJLdefeated PTAL again, if no further issue, we'll merge it.

KJLdefeated (approved):
LGTM now.

Issue #83, opened 2026-06-07
[Roadmap] RL-Kernel Roadmap Q3 - Q4


Flink-ddd:
# RL-Kernel Roadmap (2026)

We have organized RL-Kernel's goals for the second half of 2026 into several focus areas. Discussion happens on our Discord.

Repo: https://github.com/RL-Align/RL-Kernel

Discord: https://discord.com/invite/5HfkFjmPD

Hardware: NVIDIA CUDA | AMD ROCm | Ascend / Metax / Cambricon

RL-Kernel is a micro-level RL operator library. We do not own RL scheduling frameworks; verl and slime handle the macro dataflow. Our job is to replace inefficient PyTorch paths around loss, KL, log-prob, variable-length packing, and rollout/training consistency with extreme fused kernels, and to keep those kernels numerically aligned across rollout engines such as vLLM/sglang and training engines such as Megatron/DeepSpeed/FSDP.

## Roadmap Rules

Priorities are intentionally uneven:

1. P0: benchmark/CI/docs foundation, the three P0 showcase kernels, and train-inference consistency tooling.
2. P1: close one end-to-end loop and jointly validate rollout vs training consistency on top of the P0 tooling and kernels.
3. P2: large-scale kernel development across ROCm, CUDA, and Triton, including the previous fused-operator backlog.
4. P3: domestic-accelerator backend and remaining research/expansion work.

An item is considered done only when it has:

1. a code path in the repository;
2. focused tests or a reproducible smoke test;
3. a benchmark, validation report, or documented usage path when the item is user-facing;
4. clear fallback behavior for unsupported hardware or missing optional dependencies.

## Current Foundation

These pieces are already landed and form the base for the next phase:

- [x] TMA-accelerated online-softmax fused LogP for SM90+ (#26)
- [x] CUDA fused logp variants for GRPO training (#5, #33)
- [x] selected-logprob reference path for current/old/ref policy testing (#28)
- [x] minimal RL batch schema and synthetic GRPO fixtures for kernel validation (#27)
- [x] Prefix-Shared Fused Attention and same-prompt multi-output KV reuse (#65, #71)
- [x] vLLM shared-prefix rollout sampler (#34)
- [x] fused sampling op (#3)
- [x] hardware-aware context for ROCm and CUDA, plus operator HAL wrappers (#1, #36)
- [x] pure Triton FlashAttention dense fallback
[...]

PR #183, opened 2026-06-23
Add OpenAI-compatible local serving

### LINKED ISSUE #21: [FEAT][api]: implement OpenAI-compatible entry point for local serving
Build an upstream HTTP serving layer that implements OpenAI-compatible chat completion and generation endpoints, enabling third-party downstream toolchains to interface with the optimized model pipeline.

haoruilee:
Adds a lightweight OpenAI-compatible local serving entry point for RL-Kernel.

It exposes completions, chat completions, and Responses endpoints, including SSE streaming and in-memory Responses retrieval/input item support. The server can run with a local Transformers model or through `RolloutExecutor`.

A few choices are intentionally conservative: hosted-only features such as built-in tools, multimodal/audio inputs, background jobs, structured outputs, and real reasoning traces return clear OpenAI-style errors instead of being silently ignored.

Validation:
- `pre-commit run --all-files`
- `mypy --ignore-missing-imports rl_engine/`
- `python3 -m pytest tests/test_openai_server.py -q`
- `python3 -m pytest rl_engine/tests/test_dispatch.py -v`
- `mkdocs build --strict -f mkdocs.yaml`
- CUDA smoke with `sshleifer/tiny-gpt2`
- OpenAI Python SDK `base_url` create/stream/retrieve/input_items check

Closes #21

## Release Notes

* **New Features**
  * Added an OpenAI-compatible local HTTP server exposing **Completions**, **Chat Completions**, and **Responses**, including **SSE streaming** and **tool/function calling**.  
  * Added optional **in-memory Responses storage** with **create/retrieve/list/delete** endpoints.

* **Documentation**
  * Added an “OpenAI-Compatible Local Serving” user guide with supported endpoints, request/response shapes, streaming behavior, and notes on unsupported OpenAI-only features.

* **Tests**
  * Added end-to-end tests covering schemas, validation/error paths, streaming event ordering, tool calling, and storage lifecycle (plus a gated GPU smoke test).

### DIFF
--- a/rl_engine/executors/openai_server.py
+++ b/rl_engine/executors/openai_server.py
@@ -1,0 +1,2263 @@
+# SPDX-License-Identifier: Apache-2.0
+# Copyright (c) 2026 RL-Kernel Contributors
+
+from __future__ import annotations
+
+import argparse
+import json
+import os
+import threading
+import time
+import uuid
+from collections.abc import Iterator
+from dataclasses import dataclass
+from http import HTTPStatus
+from http.server import BaseHTTPRequestHandler, ThreadingHTTPServer
+from typing import Any, Mapping, NoReturn, Optional, Protocol, Sequence
+from urllib.parse import parse_qs, urlparse
+
+import torch
+
+from rl_engine.executors.rollout import RolloutExecutor
+from rl_engine.utils.logger import logger
+
+MAX_REQUEST_BODY_BYTES = 10 * 1024 * 1024
+
+
+@dataclass(frozen=True)
+class GeneratedText:
+    text: str
+    finish_reason: str
+    prompt_tokens: int = 0
+    completion_tokens: int = 0
+
+
+@dataclass(frozen=True)
+class StreamChunk:
+    text: str = ""
+    finish_reason: Optional[str] = None
+
+
+@dataclass(frozen=True)
+class StoredResponse:
+    payload: Mapping[str, Any]
+    input_items: list[Mapping[str, Any]]
+    stored_at: float
+
+
+class CompletionBackend(Protocol):
+    model_id: str
+
+    def format_chat_prompt(self, messages: Sequence[Mapping[str, Any]]) -> str: ...
+
+    def generate(
+        self,
+        prompts: Sequence[str],
+        *,
+        n: int,
+        max_tokens: int,
+        temperature: Optional[float],
+        top_p: Optional[float],
+        stop: Optional[str | Sequence[str]],
+        extra: Optional[Mapping[str, Any]] = None,
+    ) -> list[list[GeneratedText]]: ...
+
+    def stream_generate(
+        self,
+        prompt: str,
+        *,
+        max_tokens: int,
+        temperature: Optional[float],
+        top_p: Optional[float],
+        stop: Optional[str | Sequence[str]],
+        extra: Optional[Mapping[str, Any]] = None,
+    ) -> Iterator[StreamChunk | str]: ...
+
+    def count_tokens(self, text: str) -> int: ...
+
+
+class OpenAIServingError(Exception):
+    def __init__(
+        self,
+        message: str,
+        *,
+        status_code: int = HTTPStatus.BAD_REQUEST,
+        error_type: str = "invalid_request_error",
+        param: Optional[str] = None,
+    ):
+        super().__init__(message)
+        self.message = message
+        self.status_code = status_code
+        self.error_type = error_type
+        self.param = param
+
+    def to_payload(self) -> dict[str, Any]:
+        return {
+            "error": {
+                "message": self.message,
+                "type": self.error_type,
+                "param": self.param,
+                "code": None,
+            }
+        }
+
+
+class _CancelGenerationCriteria:
+    def __init__(self, stop_event: threading.Event):
+        self.stop_event = stop_event
+
+    def __call__(self, input_ids: torch.Tensor, scores: Any, **kwargs: Any) -> torch.Tensor:
+        del scores, kwargs
+        return torch.full(
+            (input_ids.shape[0],),
+            self.stop_event.is_set(),
+            device=input_ids.device,
+            dtype=torch.bool,
+        )
+
+
+class TransformersCompletionBackend:
+    """Local causal-LM backend for OpenAI-compatible development serving."""
+
+    def __init__(
+        self,
+        model_id: str,
+        *,
+        device: str = "auto",
+        dtype: str = "auto",
+        trust_remote_code: bool = False,
+    ):
+        from transformers import AutoModelForCausalLM, AutoTokenizer
+
+        self.model_id = model_id
+        self.device = _resolve_device(device)
+        self.tokenizer = AutoTokenizer.from_pretrained(
+            model_id,
+            trust_remote_code=trust_remote_code,
+        )
+        if self.tokenizer.pad_token is None:
+            self.tokenizer.pad_token = self.tokenizer.eos_token
+        self.tokenizer.padding_side = "left"
+
+        dtype_value = _resolve_dtype(dtype, self.device)
+        model_kwargs: dict[str, Any] = {"trust_remote_code": trust_remote_code}
+        if dtype_value is not None:
+            model_kwargs["dtype"] = dtype_value
+        try:
+            self.model = AutoModelForCausalLM.from_pretrained(model_id, **model_kwargs)
+        except TypeError:
+            if dtype_value is not None:
+                model_kwargs.pop("dtype", None)
+                model_kwargs["torch_dtype"] = dtype_value
+            self.model = AutoModelForCausalLM.from_pretrained(model_id, **model_kwargs)
+        self.model.to(self.device)
+        self.model.eval()
+        logger.info("Loaded %s for OpenAI-compatible serving on %s", model_id, self.device)
+
+    def format_chat_prompt(self, messages: Sequence[Mapping[str, Any]]) -> str:
+        if getattr(self.tokenizer, "chat_template", None):
+            return self.tokenizer.apply_chat_template(
+                list(messages),
+                tokenize=False,
+                add_generation_prompt=True,
+            )
+        return fallback_chat_prompt(messages)
+
+    def count_tokens(self, text: str) -> int:
+        return len(self.tokenizer.encode(text, add_special_tokens=False))
+
+    @torch.inference_mode()
+    def generate(
+        self,
+        prompts: Sequence[str],
+        *,
+        n: int,
+        max_tokens: int,
+        temperature: Optional[float],
+        top_p: Optional[float],
+        stop: Optional[str | Sequence[str]],
+        extra: Optional[Mapping[str, Any]] = None,
+    ) -> list[list[GeneratedText]]:
+        expanded_prompts = [prompt for prompt in prompts for _ in range(n)]
+        encoded = self.tokenizer(
+            expanded_prompts,
+            return_tensors="pt",
+            padding=True,
+            truncation=False,
+        )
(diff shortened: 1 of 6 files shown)


rl_engine/executors/openai_server.py
```diff
@@ -0,0 +1,1936 @@
...
+    content = item.get("content", "")
+    if isinstance(content, str):
+        text = content
+    elif isinstance(content, Sequence) and not isinstance(content, (bytes, bytearray)):
+        text_parts = []
+        for part in content:
+            if isinstance(part, str):
+                text_parts.append(part)
+            elif isinstance(part, Mapping):
+                part_type = part.get("type", "input_text")
```
inaniloquentee, line 1619: Accept "output_text" here so prior response output items can be reused as Responses input.

rl_engine/executors/openai_server.py
```diff
@@ -0,0 +1,1936 @@
...
+
+def _response_message_item(item_id: str, text: str, *, status: str) -> dict[str, Any]:
+    return {
+        "id": item_id,
+        "type": "message",
+        "status": status,
+        "role": "assistant",
+        "content": [
+            {
+                "type": "output_text",
```
inaniloquentee, line 2140: This emits "output_text", but the input parser rejects that type on the next turn.

inaniloquentee (changes requested):

rl_engine/executors/openai_server.py
```diff
@@ -0,0 +1,2263 @@
...
+            record = self._responses.get(response_id)
+        if record is None:
+            raise OpenAIServingError(
+                f"Response {response_id!r} was not found.",
+                status_code=HTTPStatus.NOT_FOUND,
+                error_type="not_found_error",
+            )
+        return record
+
+    def _evict_stored_responses_locked(self) -> None:
```
Flink-ddd, line 933: This currently iterates over the entire dictionary O(N) while holding a lock to check for TTL expiration. If self._responses is initialized as a collections.OrderedDict, entries are inherently ordered chronologically. You could just pop from the head and break upon encountering the first unexpired entry, optimizing the eviction process to O(1).

rl_engine/executors/openai_server.py
```diff
@@ -0,0 +1,2263 @@
...
+    if device.startswith("cuda") and not torch.cuda.is_available():
+        raise OpenAIServingError(
+            "CUDA was requested but torch.cuda.is_available() is false.",
+            status_code=HTTPStatus.INTERNAL_SERVER_ERROR,
+            error_type="server_error",
+        )
+    return device
+
+
+def _resolve_dtype(dtype: str, device: str) -> Optional[torch.dtype]:
```
Flink-ddd, line 2242: Hardcoding the fallback to torch.float16 for GPUs is fine, but for Ampere/Hopper or modern AMD architectures, bfloat16 is generally preferred to avoid numerical overflow. Consider detecting hardware capability here and defaulting to torch.bfloat16 where supported.

rl_engine/executors/openai_server.py
```diff
@@ -0,0 +1,2263 @@
...
+            [_CancelGenerationCriteria(stop_event)]
+        )
+        generation_kwargs["streamer"] = streamer
+
+        generation_errors: list[Exception] = []
+        generation_outputs: list[Any] = []
+
+        def _run_generate() -> None:
+            try:
+                generation_outputs.append(self.model.generate(**encoded, **generation_kwargs))
```
Flink-ddd, line 284: Capturing cross-thread exceptions using a list closure is slightly unrefined, though Python's GIL makes it thread-safe here. It works perfectly fine for this scope, but using queue.Queue or concurrent.futures.Future might be more idiomatic.

Side note: The implementation of _CancelGenerationCriteria(stop_event) to preemptively halt GPU generation when an SSE stream disconnects is an excellent, production-grade detail. Great job preventing compute leakage here.

tests/test_openai_server.py
```diff
@@ -0,0 +1,1077 @@
...
+        service.retrieve_response(first["id"])
+    assert service.retrieve_response(second["id"])["id"] == second["id"]
+    assert service.retrieve_response(third["id"])["id"] == third["id"]
+
+
+def test_response_store_evicts_expired_entries():
+    service = OpenAICompletionService(
+        RecordingBackend(),
+        response_store_max=10,
+        response_store_ttl=0.001,
```
Flink-ddd, line 512: Relying on time.sleep(0.01) with a 0.001 TTL can occasionally cause flaky tests in heavily loaded CI environments (like GitHub Actions) due to CPU starvation or thread scheduling jitter.

To make this deterministically robust, consider using unittest.mock.patch to mock time.monotonic() (which is used in _evict_stored_responses_locked), allowing you to manually advance time. Alternatively, just bumping the TTL and sleep to slightly larger values (e.g., ttl=0.1, sleep=0.2) would add more buffer against CI jitter.

tests/test_openai_server.py
```diff
@@ -0,0 +1,1077 @@
...
+                "model": [],
+                "messages": [{"role": "user", "content": "hello"}],
+                "max_tokens": 2,
+            }
+        )
+
+    assert backend.calls == []
+
+
+def test_http_endpoints_return_json():
```
Flink-ddd, line 348: There is a lot of repeated boilerplate across the HTTP tests for starting and tearing down the ThreadingHTTPServer (e.g., the try...finally: server.shutdown() blocks).

You might want to extract this into a pytest.fixture with a yield statement to handle the server lifecycle. This would significantly clean up the test bodies and ensure consistent teardown logic across all HTTP-related tests.

Flink-ddd (changes requested):
Overall, this is a high-quality PR with a clean architectural design. Implementing a lightweight OpenAI-compatible layer using the built-in ThreadingHTTPServer without pulling in heavy dependencies (like FastAPI or Uvicorn) is an elegant and pragmatic choice for local serving.

Here is the review breakdown from an infrastructure perspective:

rl_engine/executors/openai_server.py
haoruilee: Addressed. Switched the response store to OrderedDict and updated eviction to pop from the head until the first unexpired entry, then apply max-size eviction with popitem(last=False).

rl_engine/executors/openai_server.py
haoruilee: Addressed. dtype="auto" now prefers bfloat16 on ROCm and on CUDA devices where torch.cuda.is_bf16_supported(including_emulation=False) returns true, while preserving float16 fallback for older GPUs.

rl_engine/executors/openai_server.py
haoruilee: Addressed. Replaced the list-based cross-thread result/error capture with a concurrent.futures.Future, so the streaming generation thread now propagates results and exceptions through a more idiomatic synchronization primitive.

tests/test_openai_server.py
haoruilee: Addressed. Replaced the sleep-based TTL test with a deterministic monkeypatch of time.monotonic(), so the test no longer depends on scheduler timing.

tests/test_openai_server.py
haoruilee: Addressed. Added a pytest fixture for the ThreadingHTTPServer lifecycle and migrated the HTTP tests to use it, keeping shutdown/server_close/thread.join teardown consistent.

Flink-ddd:
cc @inaniloquentee @KJLdefeated PTAL again when available

KJLdefeated (commented):
Before the line-level items, a scope question I want settled first, because it determines whether the rest is even worth grinding through.
Does a bespoke OpenAI-compatible server belong in this repo? This is ~1400 lines implementing Completions, Chat, and the full Responses API — SSE framing, tool-calling, in-memory response storage — as a hand-rolled stdlib server. This repo's identity is bitwise train-inference consistency and memory-efficient kernels for GRPO/PPO; it already integrates vLLM for rollout, and vLLM ships a maintained OpenAI-compatible server (vllm serve). So I need the PR to answer, explicitly:
- What does this server do that vllm serve (or a thin adapter in front of it) doesn't? If the answer is "expose the rollout backend over an OpenAI API," that's a case for a thin wrapper over vLLM's server, not a from-scratch reimplementation.
- Who owns this long-term? The OpenAI Responses spec is large and moves fast. Re-implementing it by hand in one file is a standing maintenance liability for a kernel/training repo, and every spec change becomes our problem.
- Where does it sit in the product? Is this meant for local dev/eval, or a supported serving path? That changes the bar entirely (a dev tool tolerates gaps a network-facing service can't).

haoruilee:
Hi @KJLdefeated,

  Thanks for raising the scope question. I agree this is a critical point that needs to be settled first.

  The intended positioning of this feature is strictly an OpenAI-shaped local integration adapter for RL-Kernel executors.

  1. Where does it sit in the product?

  Its role is to provide a standard client-facing entry point for local development, eval, smoke testing, and downstream integration examples. This matches the roadmap wording in  #83: “OpenAI-compatible local serving entry point (https://github.com/RL-Align/RL-Kernel/issues/21), treated as an integration reference rather than a core serving product”

  2. What value does it add relative to `vllm serve`?

  The value is that it turns RL-Kernel’s internal executor path into a small, standard, locally testable integration surface.

  Today RL-Kernel already has internal executor components such as `RolloutExecutor`, and the rollout path can reuse vLLM under the hood. What is missing is a stable way for  existing OpenAI-shaped eval tools, local clients, and integration tests to enter that RL-Kernel path directly.

  This PR fills that gap by defining an adapter boundary:

  ```text
  OpenAI-shaped request
  -> RL-Kernel local adapter
  -> RL-Kernel backend protocol
  -> local Transformers backend / RolloutExecutor / future engine backend
```

  That is different from using a production OpenAI server as the entry point. A production server is the right serving solution for users, but it does not give RL-Kernel a small  reference surface for validating its own executor integration behavior across backends.

  This also helps future engine integrations such as sglang. Today main integrates vLLM through an in-process rollout sampler, not through vllm serve. A future sglang rollout  integration will likely face the same boundary question: how should OpenAI-shaped local eval/dev requests enter RL-Kernel’s executor path?

  With this adapter, a future SGLangCompletionBackend can implement the same RL-Kernel backend protocol and reuse the existing local request normalization, response shaping,  streaming framing, unsupported-field errors, and tests. Production users can still use the engine’s own serving stack, while RL-Kernel keeps one shared local integration  reference for validating executor adapters.

  So the purpose of this PR is to provide the missing local integration surface requested by #21 and described in #83: a lightweight reference entry point for exercising RL-Kernel   executors from standard OpenAI-shaped tooling.

  On the `“thin wrapper over vLLM’s server”` point: I agree that would be the right shape if the goal were only to expose vLLM rollout over HTTP.  The reason I do not think that is the right abstraction for this roadmap item is that the integration reference needs to sit above RL-Kernel’s backend protocol, not above one engine’s HTTP server. The local  Transformers smoke path, the RolloutExecutor path, and future engine adapters such as sglang should share the same request-normalization and test harness. A wrapper around `vllm  serve` would couple the reference path to vLLM’s server boundary and would not cover the non-vLLM local backend or future backend implementations.

  3. Who owns this long-term and what about maintenance?

  This is a very fair concern. Because this is a dev and eval integration tool, RL-Kernel should own only a documented local compatibility subset: basic text Chat andCompletions, basic Responses text output, and Responses function-call shapes needed by local eval or integration tooling.

  I also agree the current PR can read broader than this intended scope and I am open to narrowing the implementation.

 Does this match your interpretation of the roadmap scope? If so, I will update the PR to make this boundary explicit.

haoruilee:
@KJLdefeated @Flink-ddd   any updates?

PR #188, opened 2026-06-24
:feat(ws1): add NativeAttentionOp pure-PyTorch standard-softmax reference

maxiaosong1124:
## Summary

This pull request introduces a pure-PyTorch ground-truth reference implementation of standard softmax attention for the WS1 batch-invariant forward chain. It adds the `NativeAttentionOp` operator with numerical contract tests validating both batch invariance and per-dtype accuracy paths.

## Terminology

The PR employs WS1 alignment vocabulary from issue #108:

- **Axis-A (batch invariance)**: Row outputs must remain independent of batch composition, validated bitwise using `torch.equal` to ensure train/sample-time numerics stay synchronized
- **Axis-B (accuracy)**: Low-precision (bf16/fp16) forward paths computing in the input dtype, asserted within a per-dtype tolerance against an independent fp32 reference formula

## Motivation

Issue #108 establishes the numerical contract for WS1's batch-invariant forward chain. The Qwen3-8B attention block uses grouped-query attention (`Hq=32`, `Hkv=8`, group 4, `head_dim=128`, `scale=1/√128`). This PR covers the softmax attention stage: `out = softmax(Q Kᵀ · scale + masks) @ V`, computed with a hand-written naive softmax (subtract-max) over a fixed reduction order, deliberately avoiding `F.scaled_dot_product_attention` / flash whose reduction order is unspecified and would break the batch-invariance contract.

## Changes

**Operator Implementation** (`rl_engine/kernels/ops/pytorch/attention/standard_attn.py`):
- `NativeAttentionOp`: implements `softmax(Q Kᵀ · scale + masks) @ V` with `forward()` and `forward_fp32()` dual paths
- GQA via `repeat_interleave` (query head `h` → KV head `h // g`); causal offset `Skv - Sq + 1` valid for prefill (`Sq==Skv`) and decode (`Sq<Skv`); `key_padding_mask [B, Skv]` with `True`=valid; `scale` default `1/√128`
- `forward_fp32()` disables TF32 + autocast for a strict fp32 reference; pure function, fp32 accumulation, no in-place mutation

**Registry Integration** (`rl_engine/kernels/registry.py`):
- Registers `PYTORCH_NATIVE_ATTENTION` in `OpBackend`
- Routes `attention` dispatch across cuda, rocm, and cpu priority maps, distinct from the production `attn` / `PYTORCH_ATTN` SDPA fallback

**Test Suite** (`tests/test_attention.py`):
- 22 test cases covering fp32 correctness vs an independent reference, closed-form causal/decode checks, GQA replication and divisibility guard, scale defaults, key-padding masking, dtype-path accuracy, batch/chunked/padding invariance, input purity, gradient flow, registry dispatch, and a GPU-only LARGE Qwen3-8B real-shape smoke test

**Documentation** (`docs/operators/attention.md`):
- Mathematical formula, tensor contract, dual-path semantics, dispatch behavior, accuracy table, and test coverage details (plus `docs/.nav.yml` / `docs/operators/README.md` wiring)

## Contract Fulfillment

| #108 Requirement | Implementation |
|---|---|
| Deterministic reference | `forward_fp32()` with fixed-seed `torch.Generator`, fixed reduction order, no SDPA/flash |
| Per-dtype policy | Axis-A bitwise (`torch.equal`); Axis-B rel-peak tolerance (bf16 ~0.56% / 3%, fp16 ~0.07% / 0.5%) vs independent fp32 reference |
| Batch-config sweep | Full-batch computation with bitwise-identical sliced / chunked / padded row assertions |
| Realistic shapes | Tests run at Qwen3-8B attention dims (`Hq=32`, `Hkv=8`, `head_dim=128`, `scale=1/√128`) |

## Test Environment

- **OS**: Ubuntu 22.04 (kernel 5.15.0-124-generic)
- **Python**: 3.12.3
- **PyTorch**: 2.8.0+cu128
- **CUDA/cuDNN**: 12.8 / 9.10.02

## Checklist

- ✓ Pure-PyTorch reference, no custom extension required
- ✓ Hand-written naive softmax (fixed reduction order); no `F.scaled_dot_product_attention` / flash
- ✓ Covered at Qwen3-8B GQA dims (32/8/128, scale 1/√128); causal prefill + decode
- ✓ Axis-A bitwise batch invariance enforced (slice + chunked + padding)
- ✓ Axis-B per-dtype tolerance calibrated against measured drift
- ✓ Registered in OpBackend + cuda/rocm/cpu priority maps (distinct from production `attn`)
- ✓ 21 tests pass locally; 1 GPU-only LARGE smoke skips without sufficient GPU memory

## Notes

The naive path materializes the full `[B, Hq, Sq, Skv]` scores tensor (no query-chunking), so the LARGE Qwen3-8B load point (`B=8`, `Skv=4096`) is memory-heavy and GPU-only; the corresponding smoke test skips at runtime when free GPU memory is insufficient.

## Summary by CodeRabbit

* **New Features**
  * Added a PyTorch-native “standard” softmax attention operator and enabled `attention` dispatch on CPU/CUDA/ROCm, including GQA head expansion and configurable scaling.
* **Bug Fixes**
  * Improved fully-masked query-row handling to avoid NaNs and introduced a strict fp32 reference path for consistent results under autocast/TF32.
* **Documentation**
  * Added “Standard Attention” documentation and linked it in the Operators navigation.
* **Tests**
  * Added an end-to-end attention correctness/semantics test suite and updated CI to run CPU-safe attention tests by default (GPU/large cases where applicable).

### DIFF
--- a/rl_engine/kernels/ops/pytorch/attention/standard_attn.py
+++ b/rl_engine/kernels/ops/pytorch/attention/standard_attn.py
@@ -1,0 +1,182 @@
+# SPDX-License-Identifier: Apache-2.0
+# Copyright (c) 2026 RL-Kernel Contributors
+
+from __future__ import annotations
+
+import math
+from contextlib import contextmanager, nullcontext
+from typing import Optional
+
+import torch
+
+
+class NativeAttentionOp:
+    """
+    Pure PyTorch native standard-softmax attention reference.
+    out = softmax(Q Kᵀ * scale + masks) @ V
+
+    Hand-written naive softmax -- deliberately NOT
+    ``F.scaled_dot_product_attention`` / flash / mem-efficient attention, whose
+    reduction order is unspecified and would break the batch-invariance (Axis-A)
+    contract. This op defines the *correct answer* the fused kernels align to.
+
+    Qwen3-8B shapes: q ``[B, 32, Sq, 128]``, k/v ``[B, 8, Skv, 128]`` (GQA group
+    g = 32/8 = 4), scale = 1/sqrt(head_dim) = 1/sqrt(128). Heads precede seq in
+    the layout. This op covers ONLY the softmax attention; QK-Norm and RoPE are
+    applied *before* the call (see the chain test) -- the q,k passed in are
+    already normalized and rotated.
+
+    This is a reduction over the key dimension (Skv): the low-precision
+    ``forward`` path accumulates in the input dtype and therefore drifts from
+    the fp32 ``forward_fp32`` ground truth, so Axis-B accuracy uses a tolerance
+    (``torch.allclose``), not bitwise equality. Axis-A batch invariance still
+    holds bitwise within a single dtype (each query row reduces over the keys
+    independently of how many sequences share the batch).
+
+    Masking conventions:
+      * causal=True  -> upper-triangular -inf at diagonal Skv-Sq+1, valid for
+        both prefill (Sq==Skv) and decode (Sq<Skv).
+      * key_padding_mask ``[B, Skv]`` bool, True=valid / False=padding -> padded
+        key columns set to -inf (matches reference_ops.py: True=keep, False=mask).
+    """
+
+    def __init__(self) -> None:
+        """No state; the op is a pure function over (q, k, v, ...)."""
+
+    def __call__(
+        self,
+        q: torch.Tensor,
+        k: torch.Tensor,
+        v: torch.Tensor,
+        *,
+        causal: bool = True,
+        scale: Optional[float] = None,
+        key_padding_mask: Optional[torch.Tensor] = None,
+    ) -> torch.Tensor:
+        """Alias for ``forward`` so the op is callable like a module."""
+        return self.forward(q, k, v, causal=causal, scale=scale, key_padding_mask=key_padding_mask)
+
+    def forward(
+        self,
+        q: torch.Tensor,
+        k: torch.Tensor,
+        v: torch.Tensor,
+        *,
+        causal: bool = True,
+        scale: Optional[float] = None,
+        key_padding_mask: Optional[torch.Tensor] = None,
+    ) -> torch.Tensor:
+        """
+        Canonical entry: attend in the input dtype, output the input dtype.
+        This is the dtype-behavior path used as the Axis-B accuracy candidate.
+        """
+        return self._attention(
+            q,
+            k,
+            v,
+            causal=causal,
+            scale=scale,
+            key_padding_mask=key_padding_mask,
+            compute_dtype=q.dtype,
+            output_dtype=q.dtype,
+        )
+
+    def forward_fp32(
+        self,
+        q: torch.Tensor,
+        k: torch.Tensor,
+        v: torch.Tensor,
+        *,
+        causal: bool = True,
+        scale: Optional[float] = None,
+        key_padding_mask: Optional[torch.Tensor] = None,
+    ) -> torch.Tensor:
+        """Ground truth: upcast to fp32, accumulate in fp32, force fp32 output.
+
+        The whole score->softmax->value path is wrapped to disable autocast and
+        TF32 so this stays a true fp32 reference regardless of the caller's
+        ambient precision context.
+        """
+        return self._attention(
+            q,
+            k,
+            v,
+            causal=causal,
+            scale=scale,
+            key_padding_mask=key_padding_mask,
+            compute_dtype=torch.float32,
+            output_dtype=torch.float32,
+            strict_fp32=True,
+        )
+
+    # ------------------------------------------------------------------ #
+    # Helpers
+    # ------------------------------------------------------------------ #
+    @staticmethod
+    def _attention(
+        q: torch.Tensor,
+        k: torch.Tensor,
+        v: torch.Tensor,
+        *,
+        causal: bool,
+        scale: Optional[float],
+        key_padding_mask: Optional[torch.Tensor],
+        compute_dtype: torch.dtype,
+        output_dtype: torch.dtype,
+        strict_fp32: bool = False,
+    ) -> torch.Tensor:
+        """Core softmax attention: cast to ``compute_dtype``, score, mask, softmax,
+        weighted-sum over V, cast out. ``strict_fp32`` disables autocast/TF32 so the
+        fp32 reference is not silently downcast by the caller's ambient context.
+        """
+        Hq, Sq, D = q.shape[1], q.shape[2], q.shape[3]
+        Hkv, Skv = k.shape[1], k.shape[2]
+        if Hq % Hkv != 0:
+            raise ValueError(f"Hq={Hq} not divisible by Hkv={Hkv} (GQA group)")
+
+        ctx = NativeAttentionOp._strict_fp32_math(q.device.type) if strict_fp32 else nullcontext()
+        with ctx:
+            qf = q.to(compute_dtype)
+            kf = k.to(compute_dtype)
+            vf = v.to(compute_dtype)
+
+            # GQA: replicate each KV head g=Hq//Hkv times (Qwen3: 32/8 -> 4).
+            # repeat_interleave (not repeat) keeps each KV head's copies adjacent
+            # so query head h maps to KV head h // g.
+            if Hkv != Hq:
+                r = Hq // Hkv
+                kf = kf.repeat_interleave(r, dim=1)
+                vf = vf.repeat_interleave(r, dim=1)
+
+            # scale defaults to 1/sqrt(head_dim); `is not None` so an explicit 0.0 is kept.
+            scale = scale if scale is not None else (1.0 / math.sqrt(D))
+            scores = torch.matmul(qf, kf.transpose(-1, -2)) * scale  # [B, Hq, Sq, Skv]
+
+            # Causal: offset Skv-Sq+1 covers prefill (Sq==Skv) and decode (Sq<Skv).
+            if causal:
+                causal_mask = torch.triu(
+                    torch.ones(Sq, Skv, dtype=torch.bool, device=q.device),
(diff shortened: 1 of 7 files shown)


rl_engine/kernels/ops/pytorch/attention/standard_attn.py
```diff
@@ -0,0 +1,191 @@
...
+            # repeat_interleave (not repeat) keeps each KV head's copies adjacent
+            # so query head h maps to KV head h // g.
+            if Hkv != Hq:
+                r = Hq // Hkv
+                kf = kf.repeat_interleave(r, dim=1)
+                vf = vf.repeat_interleave(r, dim=1)
+
+            # scale defaults to 1/sqrt(head_dim); `is not None` so an explicit 0.0 is kept.
+            scale = scale if scale is not None else (1.0 / math.sqrt(D))
+            scores = torch.matmul(qf, kf.transpose(-1, -2)) * scale  # [B, Hq, Sq, Skv]
```
inaniloquentee, line 153: padding changes the QK/softmax shape here, so a sequence padded to Skv=10 is not bitwise-equal to the same sequence at Skv=6. The new padding test fails locally with ~1e-6 drift. We need a fixed reduction path that actually skips masked keys, or we need to relax the Axis-A bitwise claim/tests.

inaniloquentee (changes requested):

inaniloquentee:
this PR adds tests/test_attention.py, but CI never runs it. That means the new attention contract can be broken and the PR still goes green. Please add the new CPU-safe attention tests to this job.

maxiaosong1124:
@inaniloquentee thanks! Both addressed in [c194db1]:

1. CI: added a CPU-safe step running pytest tests/test_attention.py -k "not large and not gpu" (excludes the two GPU-only tests). Locally 21 passed.

2. Padding bitwise: relaxed the claim/test. Padding changes the softmax reduction width (Skv=10 vs 6), so bitwise equality isn't guaranteed in IEEE 754. Test now uses allclose(atol=1e-6), and the Axis-A bullet in the docs excludes key_padding_mask from the atol=0 claim (slicing + chunked stay bitwise).

inaniloquentee:
> @inaniloquentee thanks! Both addressed in [[c194db1](https://github.com/RL-Align/RL-Kernel/commit/c194db1f763234678945b7271b0f5c0b75a7a07f)]:
> 
> 1. CI: added a CPU-safe step running pytest tests/test_attention.py -k "not large and not gpu" (excludes the two GPU-only tests). Locally 21 passed.
> 2. Padding bitwise: relaxed the claim/test. Padding changes the softmax reduction width (Skv=10 vs 6), so bitwise equality isn't guaranteed in IEEE 754. Test now uses allclose(atol=1e-6), and the Axis-A bullet in the docs excludes key_padding_mask from the atol=0 claim (slicing + chunked stay bitwise).

Thanks for adding the CI step, that part looks good.
One issue is still not fully addressed on my side though: the new CPU-safe command still fails locally with the current padding tolerance.
I ran:

python -m pytest tests/test_attention.py -v -k "not large and not gpu"
and got:

test_key_padding_mask_excludes_padded_keys FAILED
Padding-masked result diverges from valid-only by 1.31e-06 > 1e-06

So the direction of relaxing padding from bitwise to near-equality makes sense, but atol=1e-6 looks too tight / platform-sensitive. Could you bump it with a little headroom, e.g. 2e-6 or a small peak-relative tolerance?

Also minor doc nit: the detailed Axis-A section now correctly excludes key_padding_mask from bitwise, but the test coverage line still says “slice + chunked + padding”, which can read like padding is still part of the bitwise Axis-A claim.

maxiaosong1124:
> > @inaniloquentee thanks! Both addressed in [[c194db1](https://github.com/RL-Align/RL-Kernel/commit/c194db1f763234678945b7271b0f5c0b75a7a07f)]:
> > 
> > 1. CI: added a CPU-safe step running pytest tests/test_attention.py -k "not large and not gpu" (excludes the two GPU-only tests). Locally 21 passed.
> > 2. Padding bitwise: relaxed the claim/test. Padding changes the softmax reduction width (Skv=10 vs 6), so bitwise equality isn't guaranteed in IEEE 754. Test now uses allclose(atol=1e-6), and the Axis-A bullet in the docs excludes key_padding_mask from the atol=0 claim (slicing + chunked stay bitwise).
> 
> Thanks for adding the CI step, that part looks good. One issue is still not fully addressed on my side though: the new CPU-safe command still fails locally with the current padding tolerance. I ran:
> 
> python -m pytest tests/test_attention.py -v -k "not large and not gpu" and got:
> 
> test_key_padding_mask_excludes_padded_keys FAILED Padding-masked result diverges from valid-only by 1.31e-06 > 1e-06
> 
> So the direction of relaxing padding from bitwise to near-equality makes sense, but atol=1e-6 looks too tight / platform-sensitive. Could you bump it with a little headroom, e.g. 2e-6 or a small peak-relative tolerance?
> 
> Also minor doc nit: the detailed Axis-A section now correctly excludes key_padding_mask from bitwise, but the test coverage line still says “slice + chunked + padding”, which can read like padding is still part of the bitwise Axis-A claim.

@inaniloquentee thanks for the careful follow-up! Both points addressed in 1e3a990:

1. Padding tolerance. You're right that atol=1e-6 was too tight / platform-sensitive — the drift comes from the differing softmax reduction width (Skv=10 vs 6), not from the masking semantics. Bumped _PADDING_ATOL to 2e-6, which clears the observed ~1.3e-6 with headroom. pytest tests/test_attention.py -k "not large and not gpu" → 21 passed locally, including test_key_padding_mask_excludes_padded_keys. Kept rtol=0.0 so it stays an absolute-error check.
2. Doc nit. Good catch — the coverage line did still read as if padding were part of the bitwise Axis-A claim. Updated it to slice + chunked, bitwise; padding is near-equality only, see below, so it now matches the detailed Axis-A section that excludes key_padding_mask from the atol=0 claim.

tests/test_attention.py
```diff
@@ -0,0 +1,433 @@
...
+
+def test_gradient_flows():
+    """fp32 autograd (the backward golden source) yields finite grads for q, k, v."""
+    op = NativeAttentionOp()
+    q, k, v = _qkv(2, 8, 8, seed=8)
+    q, k, v = q.requires_grad_(True), k.requires_grad_(True), v.requires_grad_(True)
+    op.forward_fp32(q, k, v, causal=True).sum().backward()
+    for t in (q, k, v):
+        assert t.grad is not None and t.grad.shape == t.shape
+        assert torch.isfinite(t.grad).all()
```
KJLdefeated: isfinite can't tell a correct gradient from a wrong-but-finite one, and attention's backward (softmax Jacobian + dQ/dK/dV contractions) is the most error-prone in the stack.
Fix: check grads against autograd through the independent double-precision reference, with a random cotangent (not .sum(), which collapses the contraction):
```python
def test_gradient_matches_reference():
    q, k, v = _qkv(2, 8, 8, seed=8)
    q, k, v = q.requires_grad_(True), k.requires_grad_(True), v.requires_grad_(True)
    dy = torch.randn_like(NativeAttentionOp().forward_fp32(q, k, v, causal=True))
    NativeAttentionOp().forward_fp32(q, k, v, causal=True).backward(dy)

    qd, kd, vd = (t.detach().double().requires_grad_(True) for t in (q, k, v))
    _ref_softmax_attn(qd, kd, vd, causal=True).backward(dy.double())
    for t, td in ((q, qd), (k, kd), (v, vd)):
        torch.testing.assert_close(t.grad, td.grad.float(), rtol=1e-4, atol=1e-4)
```

KJLdefeated (changes requested):
Overall good and clean to me. Happy to approve once the requests are fullfilled.

tests/test_attention.py
maxiaosong1124: @KJLdefeated Fixed in 674e83a — the gradient test now backprops a random (seeded) cotangent and asserts q/k/v grads against autograd through the fp64 _ref_softmax_attn, instead of just isfinite. Discriminating and TF32-immune; all 22 tests pass.

KJLdefeated (approved):
LGTM

tests/test_attention.py
```diff
@@ -0,0 +1,448 @@
...
+def _single_thread():
+    """Pin CPU GEMM to one thread so the matmul reduction order is batch-independent."""
+    prev = torch.get_num_threads()
+    torch.set_num_threads(1)
+    try:
+        yield
+    finally:
+        torch.set_num_threads(prev)
+
+
```
Flink-ddd, line 78: Please add a test_backward_batch_invariance_slice test. You should run a full-batch forward/backward pass (inside the _single_thread() context manager to lock the reduction order), then run a single-batch sliced forward/backward pass using the corresponding slice of the upstream gradient (dy), and assert that the resulting q.grad, k.grad, and v.grad slices are bitwise identical using torch.equal.

Flink-ddd (changes requested):
Aside from that, everything else is fine. If it's all sorted out, I'd be approve to merge it.

tests/test_attention.py
maxiaosong1124: @Flink-ddd  Thanks for the review! Added test_backward_batch_invariance_slice in 22d45e6.

It does exactly what you suggested:
- Full-batch forward + backward under _single_thread() to lock the reduction order
- Single-batch sliced forward + backward using the corresponding slice of the upstream gradient (dy[:1])
- Asserts the q.grad / k.grad / v.grad slices are bitwise identical to the full-batch grads via torch.equal

Parametrized over the Axis-A dtypes — all 3 pass locally.

Flink-ddd (approved):
LGTM now, Thank you for update. so fast!

Flink-ddd:
cc @inaniloquentee PTAL, if there are no further issue, we will merge it.

inaniloquentee (approved):

PR #189, opened 2026-06-24
[FEAT][kernels] Add tensor-parallel linear_logp path

inaniloquentee:
## Summary

Tensor-parallel `linear_logp` now supports vocab-sharded LM-head weights through
the public operator API. The TP path accepts global `target_ids`, local
`lm_head_weight[vocab_start:vocab_end]`, optional local bias, a TP process group,
and global vocab metadata.

| Path | Status |
| --- | --- |
| PyTorch TP path | Shared autograd implementation for native, Triton, and SM90 wrappers. |
| Registry/Triton TP wrappers | TP kwargs delegate into the shared TP path instead of local non-TP kernels. |
| SM90 TP wrapper | Constructs the SM90 op when compiled, then delegates TP kwargs to the shared TP path. |
| 4-GPU runbook | Validated on 4x H100 with NCCL. |

The TP implementation computes global log-sum-exp across vocab shards, sums the
rank-local target logit contribution, all-reduces `hidden.grad`, and keeps
`weight.grad` / `bias.grad` local to each vocab shard.

## Implementation

- Added TP metadata validation for contiguous `[0, V)` vocab partitions and
  exactly-one-owner target coverage.
- Added chunked forward stats over local vocab shards: local max, local sum,
  local target logit, and owner count.
- Added custom TP autograd:
  - forward merges local online-softmax state with NCCL collectives;
  - backward recomputes local logits by chunks;
  - `hidden.grad` is all-reduced across ranks;
  - `weight.grad` and `bias.grad` remain shard-local.
- Aligned low-precision TP math with fused/Triton bf16 semantics by computing
  bf16/fp16 matmuls in fp32 before casting returned gradients back to input
  dtypes.
- Fixed SM90 extension build issues:
  - TMA inline PTX now uses the CUDA 12.4-supported
    `shared::cluster.global.tile` form.
  - TMA/mbarrier helpers use shared-address operands accepted by ptxas.
  - The extension links with a PyTorch library rpath so `_C` imports without
    manual `LD_LIBRARY_PATH`.

## Validation Environment

| Item | Value |
| --- | --- |
| GPU | 4x NVIDIA H100 80GB HBM3 |
| Driver | 580.126.09 |
| CUDA driver/runtime | driver CUDA 13.0, PyTorch CUDA 12.4 |
| PyTorch | 2.4.1+cu124 |
| Distributed backend | NCCL, world_size=4 |
| Extension | `KERNEL_ALIGN_FORCE_SM90=1`, `_C.fused_linear_logp_sm90` available |

### Native TP math sanity

```bash
torchrun --standalone --nproc_per_node=4 scripts/test_linear_logp_tp.py \
  --op-source native \
  --dtype fp32 \
  --tokens 128 \
  --hidden-size 256 \
  --vocab-size 4096 \
  --uneven-shards
```

Result: PASS.

| Metric | max_abs | max_rel |
| --- | ---: | ---: |
| output | 4.196167e-05 | 1.038115e-06 |
| hidden_grad | 6.890297e-05 | 1.170046e-01 |
| weight_grad | 5.006790e-05 | 8.836515e-02 |
| bias_grad | 1.573563e-05 | 1.523999e-02 |

### Registry bf16 main gate

```bash
torchrun --standalone --nproc_per_node=4 scripts/test_linear_logp_tp.py \
  --op-source registry \
  --dtype bf16 \
  --reference-mode fp32 \
  --tokens 256 \
  --hidden-size 512 \
  --vocab-size 8192 \
  --uneven-shards
```

Result: PASS.

| Metric | max_abs | max_rel |
| --- | ---: | ---: |
| output | 5.340576e-05 | 1.703110e-06 |
| hidden_grad | 1.562500e-02 | 8.754433e+00 |
| weight_grad | 1.562500e-02 | 1.062290e-01 |
| bias_grad | 4.882812e-04 | 6.451613e-03 |

### Wrapper delegation

Both wrapper paths passed the same bf16/fp32-reference TP correctness run:

| op_source | Result |
| --- | --- |
| `triton` | PASS |
| `sm90` | PASS |

SM90 log confirms the extension was linked:

```text
Successfully linked to precompiled _C.fused_linear_logp_sm90 kernel.
```

### Same-dtype bf16 drift check

```bash
torchrun --standalone --nproc_per_node=4 scripts/test_linear_logp_tp.py \
  --op-source registry \
  --dtype bf16 \
  --reference-mode matching \
  --tokens 128 \
  --hidden-size 512 \
  --vocab-size 8192 \
  --atol 0.75 \
  --rtol 0.75
```

Result: PASS with recorded full-GEMM vs shard-GEMM drift.

| Metric | max_abs | max_rel |
| --- | ---: | ---: |
| output | 3.042984e-01 | 6.301940e-03 |
| hidden_grad | 8.593750e-01 | 3.489747e+02 |
| weight_grad | 5.625000e-01 | 4.345703e+06 |
| bias_grad | 1.953125e-01 | 8.200000e+00 |

This drift comes from comparing PyTorch's full bf16 GEMM against separate
vocab-shard GEMMs. The main correctness target is fp32 accumulation, matching
the fused/Triton bf16 operator semantics.

### Larger stress

```bash
torchrun --standalone --nproc_per_node=4 scripts/test_linear_logp_tp.py \
  --op-source registry \
  --dtype bf16 \
  --reference-mode fp32 \
  --tokens 256 \
  --hidden-size 512 \
  --vocab-size 8192 \
  --run-stress \
  --stress-tokens 4096 \
  --stress-hidden-size 2048 \
  --stress-vocab-size 32768
```

Result: PASS.

| Metric | Value |
| --- | ---: |
| finite | PASS |
| max_rank_elapsed_ms | 105.494 |
| max_rank_peak_memory_gb | 0.469 |

### Unit coverage

```bash
python -m pytest tests/test_linear_logp.py -q -rs
```

Result: 21 passed in 8.40s.

## Latest PR 189 Test Results

Validated on 2026-06-27 UTC on the `pr-189` / `feat/fused-linear-logp` branch.
Runtime code was validated at `ddf65a7`; follow-up `355a2e0` only removes local
PR notes from the submitted branch and passed docs/pre-commit checks.

### Environment

| Item | Value |
| --- | --- |
| Python | 3.11.10 |
| PyTorch | 2.4.1+cu124 |
| CUDA runtime/toolkit | 12.4 / `nvcc` 12.4.131 |
| Driver | 580.126.09 |
| GPU | 2 x NVIDIA H100 80GB HBM3, compute capability 9.0 |
| DeepSpeed | 0.19.2, installed with `DS_BUILD_OPS=0` |
| Extension | Editable rebuild with `KERNEL_ALIGN_FORCE_SM90=1`; `_C.fused_linear_logp_sm90` available |

### CI and Regression Checks

| Area | Command | Result |
| --- | --- | --- |
| PR-focused linear logp unit tests | `python -m pytest tests/test_linear_logp.py -q -rs` | 27 passed |
| DeepSpeed worker contract tests | `python -m pytest tests/test_deepspeed_training_worker.py -q -rs` | 22 passed |
| Fused logp fallback and CUDA loss-step regression | `python -m pytest tests/test_op_accuracy.py tests/test_rl_kernel_loss_step.py -q -rs --tb=short` | 21 passed |
| Full H100/SM90 pytest suite | `python -m pytest tests rl_engine/tests -q -rs --tb=short` | 283 passed, 82 skipped |
| CI dispatch baseline | `python -m pytest rl_engine/tests/test_dispatch.py -v` | 5 passed |
| CI attention baseline | `PYTEST_DISABLE_PLUGIN_AUTOLOAD=1 python -m pytest tests/test_attention_correctness.py -q -rs` | 45 passed, 82 skipped |
| Type check | `python -m mypy --ignore-missing-imports rl_engine/` | Success: no issues found in 50 source files |
| Documentation build | `mkdocs build --strict -f mkdocs.yaml` | Passed |
| Pre-commit | `pre-commit run --all-files` | Passed |

The attention skips were expected for this CUDA/H100 environment: external
`flash_attn` / `_C.flash_attn_forward` was unavailable, and ROCm cases were
skipped on NVIDIA CUDA.

### Production-Like CUDA Checks

| Scenario | Result | Notes |
| --- | --- | --- |
| 2-GPU direct SM90 `linear_logp` TP stress, bf16/fp32 reference, uneven shards | PASS | Stress finite check passed; max rank elapsed 184.532 ms; peak memory 1.696 GiB |
| 2-GPU registry-dispatched `linear_logp` TP, bf16/fp32 reference, uneven shards | PASS | Registry selected the fused SM90 TMA linear-logp path |
| Real DeepSpeed CUDA smoke | PASS | `DeepSpeedEngine`, zero stage 0, `current_logp_backend=FusedLinearLogpSM90Op`, finite loss, published weight version 11 |

Direct SM90 TP correctness metrics: output max abs `3.051758e-04`, hidden grad
`6.250000e-02`, weight grad `6.250000e-02`, bias grad `1.562500e-02`.
Registry TP correctness metrics: output max abs `1.068115e-04`, hidden grad
`3.125000e-02`, weight grad `3.125000e-02`, bias grad `1.953125e-03`.

### Regression Found and Fixed

A full-suite H100 run exposed that the legacy SM90 `logp` registry path selected
`FusedLogpSM90Op` for cases it did not fully support: fp16 logits, `.out(...)`,
and `.apply_fp32(...)`. The wrapper now delegates to `FusedLogpGenericOp` by
default, and the legacy SM90 TMA `logp` fast path is gated behind
`RL_KERNEL_ENABLE_EXPERIMENTAL_SM90_LOGP=1`. This does not affect the PR's fused
SM90 `linear_logp` production path, which is covered by the two-GPU checks above.

## Summary by CodeRabbit

* **New Features**
  * Added vocab-sharded tensor-parallel support for `linear_logp`, including new `tp_group`, `vocab_start_index`, and `global_vocab_size` options with TP-aware dispatch (including SM90 when eligible).
  * Added distributed scripts/runbooks and enhanced the training worker to compute log-probabilities via `linear_logp` with TP metadata.

* **Bug Fixes**
  * Improved tensor-parallel metadata validation and tightened correctness for forward/backward outputs and gradients (including ignore-index handling).

* **Documentation**
  * Expanded `linear_logp` operator docs and added an end-to-end TP test runbook.

* **Tests**
  * Added comprehensive correctness, gradient, and multi-rank TP coverage (single- and distributed).


docs/operators/linear-logp-tp-test.md
```diff
@@ -0,0 +1,303 @@
...
+  --hidden-size 512 \
+  --vocab-size 8192 \
+  --run-stress \
+  --stress-tokens 4096 \
+  --stress-hidden-size 2048 \
+  --stress-vocab-size 32768
+```
+
+Expected: `finite=PASS`. Record `max_rank_elapsed_ms` and
+`max_rank_peak_memory_gb`.
```
KJLdefeated: Non-blocking. Can you also document the benchmark result (e.g. runtime) here?

csrc/cuda/fused_linear_logp_sm90.cu
```diff
@@ -85,12 +85,14 @@ __global__ void fused_linear_logp_sm90_kernel(const __grid_constant__ CUtensorMa
...
-    int mbar[STAGES];
+    const uint64_t sH_base_tma = __cvta_generic_to_shared(sH);
+    const uint64_t sW_base_tma = __cvta_generic_to_shared(sW);
+    const uint32_t sH_base = static_cast<uint32_t>(sH_base_tma);
+    const uint32_t sW_base = static_cast<uint32_t>(sW_base_tma);
+    uint64_t mbar[STAGES];
 #pragma unroll
     for (int s = 0; s < STAGES; ++s)
-        mbar[s] = static_cast<int>(__cvta_generic_to_shared(mbar_base + 2 * s));
+        mbar[s] = __cvta_generic_to_shared(mbar_base + 2 * s);
```
KJLdefeated, line 97: I think these lines are for CUDA12.4 fix? Because some features in 12.8 and 12.4 are really different, the code dev in 12.8 may have problem in 12.4. When I previously developed linear-logp cuda kernel, I was using cuda 12.4 but can not compile some features in tma_utils.cuh by @Flink-ddd . I found the previous works were done in CUDA12.8, so I later switch in CUDA 12.8, which you can see the #Performance section in docs/operators/linear-logp.md.
I think we should stick our CUDA version to 12.8? Maybe we should discuss on this aligness.

tests/linear_logp_tp.py
KJLdefeated, line 1: May I ask why not put this test into test/ directory?

docs/operators/linear-logp-tp-test.md
```diff
@@ -0,0 +1,303 @@
...
+## What To Attach To The PR
+
+For the PR update, paste:
+
+- command lines for each run;
+- the `[env]` block;
+- all `[correctness]` metric lines;
+- the `[stress]` finite, elapsed time, and peak memory lines;
+- GPU model and CUDA/PyTorch versions;
+- any matching-reference drift numbers if that optional run fails or requires loose tolerances.
```
KJLdefeated: Non-blocking. Are these lines necessary for this doc?

KJLdefeated (commented):
Excellent work! I just have queries about dev CUDA version and documents.

docs/operators/linear-logp-tp-test.md
inaniloquentee: > Non-blocking. Can you also document the benchmark result (e.g. runtime) here?

Added the 4x H100 stress result here: 105.494 ms max-rank runtime and 0.469 GB peak memory. Thanks!

csrc/cuda/fused_linear_logp_sm90.cu
inaniloquentee: > I think these lines are for CUDA12.4 fix? Because some features in 12.8 and 12.4 are really different, the code dev in 12.8 may have problem in 12.4. When I previously developed linear-logp cuda kernel, I was using cuda 12.4 but can not compile some features in tma_utils.cuh by @Flink-ddd . I found the previous works were done in CUDA12.8, so I later switch in CUDA 12.8, which you can see the #Performance section in docs/operators/linear-logp.md. I think we should stick our CUDA version to 12.8? Maybe we should discuss on this aligness.

Good catch. This is about keeping the mbarrier shared-address operand width explicit, not changing the CUDA baseline. I added a short comment there. Happy to align separately on whether we want CUDA 12.8 as the standard perf baseline.

tests/linear_logp_tp.py
inaniloquentee: > May I ask why not put this test into test/ directory?

Moved it under tests/ as tests/linear_logp_tp.py.

docs/operators/linear-logp-tp-test.md
inaniloquentee: > Non-blocking. Are these lines necessary for this doc?

Agreed, that was more PR-process than operator docs. Removed it.

csrc/utils/tma_utils.cuh
```diff
@@ -51,33 +52,39 @@ inline void init_tensor_map(
...
-__device__ inline void tma_2d_g2s(int dst_smem_addr, const void *tmap_ptr, int x, int y, int mbar_addr) {
-    asm volatile("cp.async.bulk.tensor.2d.shared::cta.global.mbarrier::complete_tx::bytes "
+__device__ inline void tma_2d_g2s(
+    uint64_t dst_smem_addr,
+    const void *tmap_ptr,
+    int x,
+    int y,
+    uint64_t mbar_addr
+) {
+    asm volatile("cp.async.bulk.tensor.2d.shared::cluster.global.tile.mbarrier::complete_tx::bytes "
```
Flink-ddd, line 87: This line changes the TMA destination from shared::cta.global.mbarrier to shared::cluster.global.tile.mbarrier. That's a scope/semantics change, not just a CUDA 12.4 syntax fix and we're not doing a cluster launch here. Can you confirm why this is correct in the non-cluster case? We ran the SM90 path on 8x H100 SXM and results looked fine, so I'm fairly confident, but I'd like the reasoning documented in this thread before merge.

rl_engine/kernels/ops/pytorch/loss/linear_logp.py
```diff
@@ -3,14 +3,381 @@
...
+
+
+def _tensor_parallel_world_size(tp_group: Any) -> int:
+    if tp_group is None:
+        return 1
+    dist = _require_distributed_initialized()
+    return dist.get_world_size(group=tp_group)
+
+
+def should_use_tensor_parallel_linear_logp(
```
Flink-ddd, line 58: When an explicit tp_group is passed but world_size == 1, the logic appears to fall through to return True and take the TP path with a single rank. Is that intended? If not, please guard it, either way a unit test covering this case would help.

rl_engine/kernels/ops/cuda/loss/logp.py
```diff
@@ -17,14 +19,75 @@ def __init__(self):
...
-        assert logits.dtype == torch.bfloat16, "TMA logp currently requires bfloat16 logits"
-        assert logits.is_contiguous(), "Logits must be contiguous for TMA block loading"
+        return self.apply(logits, labels)
+
+    def _fallback_op(self):
+        if self._fallback is None:
+            self._fallback = FusedLogpGenericOp()
+        return self._fallback
+
+    def _can_use_sm90(self, logits: torch.Tensor) -> bool:
```
Flink-ddd, line 33: The RL_KERNEL_ENABLE_EXPERIMENTAL_SM90_LOGP gate and the default-to-FusedLogpGenericOp behavior have no test coverage. Could you add a regression test that locks in "default uses generic, env var enables SM90"? Easy to silently break otherwise.

Flink-ddd (changes requested):
Excellent work! Left a few inline comments. Happy to approve once #1 (the TMA PTX scope change) is answered. #2 and #3 can be quick follow-ups if you prefer, but let's at least track them.

csrc/cuda/fused_linear_logp_sm90.cu
Flink-ddd: Makes sense, the explicit operand width is fine.Makes sense, the explicit operand width is fine. Thanks for clarifying. Agreed on aligning separately on whether CUDA 12.8 should be our standard perf baseline, let's track that as its own discussion.

Flink-ddd (approved):
cc @KJLdefeated If there are no further questions, Let's merge this PR first. we need  base on this PR run benchmark. Thank you.

KJLdefeated (approved):
Approved

Issue #148, opened 2026-06-17
[WS1][kernels] Batch-invariant logprob (selected, locked reduction)


Flink-ddd:
Part of WS1 — Full Batch-Invariant Forward Chain (epic: #<WS1 tracking issue>)

## Why

Logprob is the last op in the chain and the quantity RL training actually compares between rollout and training — any batch drift that survives to here directly produces KL spikes and training collapse. A selected-logprob reference path already exists (#28); kernel work is in progress (#96). This issue brings that kernel under the #108 numerical contract and confirms the selected-logprob reduction is locked.

## Scope

Finalize the selected-logprob computation as batch-invariant under the shared harness.

- Input: `logits [num_tokens, vocab_size]`, `target_ids [num_tokens]`. Output: `selected_logprob [num_tokens]`.
- Computation: `selected_logprob[t] = logits[t, target_ids[t]] - logsumexp(logits[t, :])`.
- Fixed row-wise max and sum-exp reduction order; FP32 log-sum-exp accumulation over the vocab dimension; no `atomicAdd`.
- Masking / padding / ignore-index semantics must not introduce batch-dependent branches.
- Wire the kernel through the #108 harness (it is named there as the end-to-end smoke test) and validate across the standard sweep; numerics checked against the #28 reference path.

Optional extensions (separate PRs):
- `ignore_index`, token mask, entropy output, KL helper, packed-sequence support.

## Out of scope

- Re-deriving the kernel from scratch — extend #96 rather than fork it.
- Reward / advantage computation downstream of logprob.
- Multi-GPU vocab-parallel logprob.
- FP8.

## Acceptance criteria

- Selected-token logprobs are bitwise-identical for the same sequence across batch size, batch position, mixed batches, and padding layouts.
- Repeated runs are deterministic; the kernel uses no `atomicAdd`.
- Matches the #28 reference path (PyTorch `log_softmax + gather`) within #108 tolerance.
- The logprob backward path, if exposed for training through logits, passes the shared gradient-invariance check from the WS1 backward-consistency issue.
- Passes the #108 shared test helper and serves as the harness's wired-in smoke test.

## Notes

- Depends on #108; builds on #28 (reference path) and #96 (kernel — confirm the exact issue/PR number in the repo).
- This is the chain's exit signal — prioritize alongside the upstream LM-head matmul, since drift there lands here first.

## Planned PRs

[...]

PR #199, opened 2026-06-28
 [WS1][kernels] Batch-invariant logprob (Native, Triton)

hihaluemen:
## Summary

Batch-invariant selected log-prob operator with a PyTorch native reference and Triton forward/backward implementation.

This implements `batch_invariant_logp` for materialized logits:

```text
selected_logp[row] = logits[row, target_ids[row]] - logsumexp(logits[row, :])
```

The goal is not only correctness against `log_softmax + gather`, but stable row-local behavior: the same row must produce the same selected log-prob when evaluated alone, at another batch position, or with different neighboring rows.

Refs: https://github.com/RL-Align/RL-Kernel/issues/148

| Backend | Status |
| --- | --- |
| CUDA / ROCm (Triton) | Online-softmax forward with fixed vocab tiling and locked row-local reduction order; tile-wise Triton backward using saved per-row `lse`. |
| PyTorch native | FP32 row-wise reference implementation; CPU fallback and Triton-less fallback. |

Current dispatch:

```text
CUDA / ROCm: Triton -> PyTorch
CPU:         PyTorch
```

A native CUDA backend and operator benchmarks are intentionally left for a follow-up PR.

### PyTorch `NativeBatchInvariantLogpOp`

- Reshapes `[*lead, V]` logits to `[N, V]`.
- Computes selected log-probs row-wise in FP32.
- Returns an `[*lead]` float32 output.
- Supports CPU and CUDA tensors.
- Supports `ignore_index`; ignored rows output `0.0`.
- Validates non-ignored target ids by default.
- Allows validation to be disabled explicitly for trusted inputs.

### Triton `TritonBatchInvariantLogpOp`

- Forward uses one Triton program per token row.
- Vocab is scanned with fixed `_BLOCK_V=1024`; no autotune is used.
- Uses online logsumexp with a fixed left-to-right vocab-tile order.
- Saves per-row `lse` for backward.
- Backward uses one Triton program per `(row, vocab_tile)`.
- Computes:

  ```text
  grad_logits[row, j] =
      grad_out[row] * (1[j == target] - exp(logits[row, j] - lse[row]))
  ```

- Computes gradient arithmetic in FP32 and lets `tl.store` cast to the input dtype on the final write.
- Ignored rows receive zero gradient across the full vocab.
- Uses no atomic writes and no cross-row reductions.
- Target-range validation is opt-in with `validate=True`; the default hot path avoids blocking CUDA synchronization and temporary boolean-mask allocation.

### Registry and documentation

- Adds `batch_invariant_logp` to `KernelRegistry`.
- Adds operator documentation under `docs/operators/batch-invariant-logp.md`.
- Adds the operator to `docs/operators/README.md` and `docs/.nav.yml`.

## Batch-invariance contract

Each output row depends only on:

```text
logits[row, :]
target_ids[row]
ignore_index
grad_out[row]    # backward only
```

It must not depend on:

```text
batch size
row position
neighboring rows
padding layout
```

The Triton path enforces this by:

- using `grid=(num_tokens,)` for forward;
- using fixed `_BLOCK_V=1024`;
- scanning vocab tiles in deterministic left-to-right order;
- using `grid=(num_tokens, vocab_tiles)` for backward;
- reusing the saved per-row `lse` in backward;
- avoiding atomics and shared cross-row writes.

## Tensor contract

```python
from rl_engine.kernels.registry import kernel_registry

batch_invariant_logp = kernel_registry.get_op("batch_invariant_logp")

out = batch_invariant_logp(
    logits,       # [B, T, V] or [N, V], differentiable
    target_ids,   # [B, T] or [N], integer
    ignore_index=-100,
)                # -> [B, T] or [N], float32
```

Rules:

- `target_ids.shape == logits.shape[:-1]`.
- Output shape equals `target_ids.shape`.
- Output dtype is `float32`.
- Non-ignored target ids are required to be in `[0, vocab_size)`.
- PyTorch validates target ranges by default.
- Triton validation is opt-in with `validate=True` to avoid CUDA synchronization in the default hot path.
- `target_ids[row] == ignore_index` means:

  ```text
  out[row] = 0.0
  grad_logits[row, :] = 0.0
  ```

## Correctness and tests

Validated on Windows/CPU and WSL/Linux with CUDA + Triton:

```bash
python -m pytest tests/test_operator_inputs.py -q
# 16 passed

python -m pytest tests/test_batch_invariant_logp.py -q -rs
# 53 passed

python -m pytest \
  tests/test_linear_logp.py \
  tests/test_op_accuracy.py \
  tests/test_batch_invariant_logp.py \
  -q -rs
# 87 passed, 11 skipped
```

The skipped cases require Hopper/SM90 hardware and an extension built with the corresponding SM90 symbols. They are unrelated to `batch_invariant_logp`.

### Test coverage

- Correctness against `torch.log_softmax(...).gather(...)`.
- PyTorch native against the existing native logp reference.
- Triton against the PyTorch native implementation.
- fp32, fp16, and bf16 inputs.
- Float32 output dtype.
- Leading-shape preservation, including `[B, T, V] -> [B, T]`.
- Large-vocab smoke tests.
- Batch size 1 versus batch N invariance.
- The same row at different batch positions.
- Mixed neighboring batch contents.
- Repeated-run determinism.
- Padding and ignore-index layout invariance.
- Backward correctness.
- Gradient batch invariance.
- Ignored-row zero gradient.
- Invalid-target validation for negative and out-of-range targets.
- Registry dispatch.
- GTest input construction and operator registration.

## GTest / real-value validation

`batch_invariant_logp` is registered in the gtest operator checker:

- PyTorch `NativeBatchInvariantLogpOp.apply` is the gold path.
- PyTorch and Triton implementations are registered as candidates.
- Forward outputs and `logits` gradients are compared with `--check-grad`.
- The existing `logprob` tolerance contract is reused.
- Random upstream gradients are used by default for a stricter vector-Jacobian-product comparison.

### Qwen3 shape matrix

Tests use the Qwen3 vocabulary size `V = 151936` with:

- dtypes: `fp32`, `bf16`, `fp16`;
- batch sizes: `B = (1, 8, 16, 32, 64)`;
- sequence lengths: `S = (1, 31, 33, 127, 129, 255, 256, 257, 512, 1024, 4096, 8192)`;
- executed cases: `B * S <= 4097`;
- backward enabled with random upstream gradients.

This results in 34 `(B, S)` shapes and 102 executed cases per platform.

| Platform | Candidate | Result |
| --- | --- | --- |
| Windows / CPU | PyTorch | 102 passed, 0 failed, 78 memory-guarded skips |
| WSL / CUDA | Triton | 102 passed, 0 failed, 78 memory-guarded skips |

The Windows run validates gtest registration, input construction, and forward/backward plumbing. The WSL run performs the real numerical comparison between the Triton candidate and the PyTorch gold path for both forward values and backward gradients.

### Example commands

PyTorch CPU:

```bash
python scripts/check_operator.py \
  --op batch_invariant_logp \
  --candidate pytorch \
  --device cpu \
  --dtype fp32 \
  --batch 1 \
  --seq 31 \
  --vocab 151936 \
  --check-grad
```

Triton CUDA:

```bash
python scripts/check_operator.py \
  --op batch_invariant_logp \
  --candidate triton \
  --device cuda \
  --dtype bf16 \
  --batch 8 \
  --seq 257 \
  --vocab 151936 \
  --check-grad
```


rl_engine/kernels/ops/triton/loss/batch_invariant_logp.py
```diff
@@ -0,0 +1,244 @@
...
+                f"logits leading shape {tuple(logits.shape[:-1])} must match "
+                f"target_ids shape {tuple(target_ids.shape)}"
+            )
+
+        vocab_size = logits.size(-1)
+        target_flat = target_ids.reshape(-1)
+        valid_targets = target_flat[target_flat != ignore_index]
+        if valid_targets.numel() > 0 and (
+            (valid_targets < 0).any() or (valid_targets >= vocab_size).any()
+        ):
```
KJLdefeated: On CUDA, .any() inside a Python if calls __bool__ → .item() → a blocking sync — twice per forward — plus the boolean-mask index allocates a tensor each call. This op runs every training step, so every step eats two stream stalls just to validate, which undercuts the point of a fast batch-invariant kernel.
I would suggest it to be opt-in / debug only:
```python
def apply(self, logits, target_ids, ignore_index=-100, *, validate=False):
    ...
    if validate:  # opt-in / debug only — this path syncs
        valid = target_flat[target_flat != ignore_index]
        if valid.numel() and ((valid < 0).any() or (valid >= vocab_size).any()):
            raise ValueError(...)
```

tests/test_triton_batch_invariant_logp.py
```diff
@@ -0,0 +1,352 @@
...
+# Copyright (c) 2026 RL-Kernel Contributors
+
+"""Tests for the Triton batch-invariant selected-logprob kernel (issue #148).
+
+These tests validate that the Triton kernel produces results that:
+1. Match the PyTorch reference implementation (correctness).
+2. Are bitwise identical across different batch sizes / positions (batch-invariance).
+3. Support backward pass (gradient correctness).
+
+All tests are skipped when Triton or CUDA is unavailable (e.g. on Windows or CPU-only).
```
KJLdefeated: Can you combine all different backend method of batch invariant logp (Triton / Torch) into one file (tests/test_batch_invariant_logp.py) for codebase cleanliness and reusable functions? For future CUDA batch-invariant logp, I will also put CUDA test in tests/test_batch_invariant_logp.py.

docs/operators/batch-invariant-logp.md
```diff
@@ -0,0 +1,172 @@
...
+
+out = op(logits, target_ids, ignore_index=-100)
+assert out.shape == target_ids.shape
+assert out.dtype == torch.float32
+assert out[0, 0].item() == 0.0
+
+out.sum().backward()
+```
+
+## Tests
```
KJLdefeated, line 154: Just make a note that this kernel is currently lacking benchmarks, I will support it in another PR (cuda ver. for this).

KJLdefeated (changes requested):
Overall correct and clean. The forward is one program per row, fixed _BLOCK_V=1024, no autotune, a sequential left-to-right online-LSE over vocab tiles, so the per-row reduction order is locked and independent of batch size, position, and neighbors. The backward is the FlashAttention pattern done right: save per-row lse in forward, recompute softmax = exp(logits − lse) in backward against that saved lse, grad = grad_out·(onehot − softmax), grid (rows, vocab_tiles), no atomics, no cross-row coupling, so the gradient is batch-invariant too, and consistent with the forward's exact lse.

Some notes:
This PR supports Batch-invariant logprob (Native / Triton), not yet CUDA (cc. me). Can you note that in PR's title? --> add Batch-invariant logprob (Native, Triton)
When this PR is done, I will make another PR to support CUDA version of this.

docs/operators/batch-invariant-logp.md
hihaluemen: Addressed, thanks.
  Changes:
  - Renamed PR scope/title to Batch-invariant logprob (Native, Triton); CUDA + benchmarks will be handled in a follow-up PR.
  - Made target range validation opt-in via validate=False by default to avoid CUDA stream sync in the Triton hot path.
  - Merged Native and Triton tests into tests/test_batch_invariant_logp.py.
  - Updated operator docs to mention validate=True and benchmark/CUDA follow-up.

rl_engine/kernels/ops/triton/loss/batch_invariant_logp.py
```diff
@@ -0,0 +1,247 @@
...
+            num_tokens,
+            vocab_size,
+            logits_2d.stride(0),
+            ignore_index=ignore_index,
+            BLOCK_V=_BLOCK_V,
+        )
+
+        grad_logits = grad_logits.to(logits_2d.dtype).reshape(
+            ctx.lead_shape + (vocab_size,)
+        )
```
KJLdefeated: Backward allocates a full fp32 [N, V] grad buffer, then casts an extra [N,V] allocation + a whole HBM read/write pass. line163: empty_like(..., dtype=torch.float32) then line179: .to(logits_2d.dtype). For a bf16 model with large V this doubles the gradient memory transiently (fp32 [N,V] + bf16 [N,V]) and adds a full extra pass. 
Fix: allocate grad_logits = torch.empty_like(logits_2d) (input dtype) and let tl.store downcast the fp3 computed grad on write, saves the fp32 buffer and the cast pass.

KJLdefeated (changes requested):
Previous changes are good to me. I found some logic in triton kernel can be improved. After fix the requests, I am happy to approve.

rl_engine/kernels/ops/triton/loss/batch_invariant_logp.py
hihaluemen: Addressed: Triton backward now allocates grad_logits in the input dtype and lets tl.store cast on write, so we keep fp32 math in-kernel without the extra fp32 buffer/cast pass; I reran tests/test_batch_invariant_logp.py on WSL and it passes.

KJLdefeated (approved):
LGTM now. Thx for great work. Approved.

Flink-ddd:
Hi @hihaluemen , this PR need to verify and pass the ground truth framework before I start the review process. Thank you.

hihaluemen:
GPU CI reaches the full test suite successfully. The only failure appears unrelated to this PR:

```text
FAILED tests/test_matmul.py::
TestNativeMatmulOpBatchInvariance::test_batch_grad_invariance
1 failed, 605 passed, 102 skipped
```

The failing test is already present on the base branch and uses CPU tensors. It compares full-batch versus per-row `torch.matmul` input gradients with `torch.equal`. Limiting PyTorch to one CPU thread does not guarantee bitwise GEMM equality across different matrix shapes or kernel dispatches.

All `batch_invariant_logp` tests passed, including the PyTorch/Triton forward and backward gtest comparisons. Could a maintainer confirm whether this base matmul test should use a numerical tolerance or a separately implemented deterministic matmul path?

Flink-ddd:
cc @a-kaa PTAL, I think I've identified the root cause that is currently blocking the GPU CI across all recent kernel PRs. The consistent failure comes from test_matmul.py::TestNativeMatmulOpBatchInvariance::test_batch_grad_invariance.

Currently, this test uses torch.equal() to compare gradients. However, the native PyTorch/cuBLAS matmul does not guarantee bitwise identical results across different batch splits due to varying floating-point accumulation orders. This naturally leads to precision drift ($\sim 10^{-4}$), causing a flaky test.

To unblock the CI for everyone, we should relax this specific assertion from strict bitwise equality to numerical tolerance.

**Change from:**
`assert torch.equal(full_a.grad, torch.stack(single_a_grads))`

To:

`torch.testing.assert_close(full_a.grad, torch.stack(single_a_grads), rtol=1e-4, atol=1e-4)`

Once this minor test fix is merged into the base branch, all pending operator PRs can merge the latest main and successfully pass the CI validation.

@hihaluemen Could you please help investigate this issue and submit a bugfix?

hihaluemen:
> cc @a-kaa PTAL, I think I've identified the root cause that is currently blocking the GPU CI across all recent kernel PRs. The consistent failure comes from test_matmul.py::TestNativeMatmulOpBatchInvariance::test_batch_grad_invariance.
> 
> Currently, this test uses torch.equal() to compare gradients. However, the native PyTorch/cuBLAS matmul does not guarantee bitwise identical results across different batch splits due to varying floating-point accumulation orders. This naturally leads to precision drift ( ∼ 10 − 4 ), causing a flaky test.
> 
> To unblock the CI for everyone, we should relax this specific assertion from strict bitwise equality to numerical tolerance.
> 
> **Change from:** `assert torch.equal(full_a.grad, torch.stack(single_a_grads))`
> 
> To:
> 
> `torch.testing.assert_close(full_a.grad, torch.stack(single_a_grads), rtol=1e-4, atol=1e-4)`
> 
> Once this minor test fix is merged into the base branch, all pending operator PRs can merge the latest main and successfully pass the CI validation.
> 
> @hihaluemen Could you please help investigate this issue and submit a bugfix?

Got it, I’ll investigate this issue. Thanks for identifying the likely root cause.

maxiaosong1124:
LGTM!

maxiaosong1124 (approved):

Flink-ddd (approved):
LGTM! Thanks.

PR #208, opened 2026-07-05
[Feat][Kernel] Add TP Fused linear logp Triton

KJLdefeated:
# [FEAT][kernels] Add tensor-parallel path to the Triton linear_logp kernel

## Summary

Gives the Triton `linear_logp` op a real vocab-parallel (tensor-parallel)path,  so a vocab-sharded LM head runs the fused Triton forward on each rank instead of silently falling back to the pure-PyTorch TP implementation.

```text
logp[t] = (hidden @ W_local^T + b_local)[t, target[t]] - logsumexp_over_all_ranks(...)
```

| Path | Status |
| --- | --- |
| Triton TP forward (`_TensorParallelTritonLinearLogpFunction`) | Local-shard fused forward + cross-rank logsumexp merge. New. |
| Triton op dispatch | `TritonLinearLogpOp` now routes TP calls. |
| Shared TP backward (`tensor_parallel_chunked_linear_logp_backward`) | Extracted from the byte-identical native + SM90 backwards; now used by all three. |
| Tests | `tests/linear_logp_tp.py --op-source triton` |

Builds on the merged #122 (fused linear_logp) and #189 (native + SM90 TP).

### Triton TP forward — `rl_engine/kernels/ops/triton/loss/linear_logp.py`
- Factored the forward-kernel launch into `_run_forward_kernel(...) -> (logp, lse)`,
  reused by both the plain and TP forwards (no kernel change).
- `_TensorParallelTritonLinearLogpFunction` mirrors the SM90 TP path
  (`_FusedTensorParallelLinearLogpSM90Function`):
  - each rank clamps the global target to a valid **local** column and runs the
    fused forward over its shard, giving the local `lse` and (for the owned
    token) the target logit, recovered as `target_logit = logp + lse`;
  - ranks merge shards with the standard online-softmax reduction:
    `all_reduce(MAX)` the running max, rescale + `all_reduce(SUM)` the running
    sum, `all_reduce(SUM)` the target logit; `logp = target_logit - global_lse`;
  - `_validate_tp_vocab_partition` + `_validate_global_targets` (from #189)
    guarantee a contiguous `[0, V)` partition and in-range targets, so each
    target has exactly one owner — matching the SM90 path (no extra owner-count
    all-reduce).
- `TritonLinearLogpOp.apply` now dispatches TP calls to
  `_triton_tensor_parallel_linear_logp`.

### Shared TP backward — `rl_engine/kernels/ops/pytorch/loss/linear_logp.py`
- The native and SM90 TP backwards were byte-identical. Reuse`tensor_parallel_chunked_linear_logp_backward`.

## Validation

Validated base on `torchrun --standalone --nproc_per_node=4 tests/linear_logp_tp.py --op-source triton` reference from [docs/operators/linear-logp-tp-test.md](https://github.com/RL-Align/RL-Kernel/blob/main/docs/operators/linear-logp-tp-test.md)

### Results (backend=nccl, world_size=4)

| run | result |
| --- | --- |
| triton **fp32** (`--atol 1e-3 --rtol 1e-3`) | **PASS** — output 7.63e-6, all grads pass |
| triton **bf16** correctness | output PASS (2.69e-1); grads = native/sm90 (see below) |
| triton **bf16** uneven shards | output PASS |
| triton **bf16** stress (tokens=4096, hidden=2048, V=32768) | **finite PASS**, 31 ms, 0.47 GB/rank |

**Backend equivalence (bf16, same harness/config, 4 NCCL ranks):**

| backend | output | hidden_grad max_abs | hidden_grad max_rel |
| --- | --- | --- | --- |
| **triton (this PR)** | PASS 2.69e-1 | 5.781250e-01 | 2.281e+02 |
| native (#189) | PASS 2.69e-1 | 5.781250e-01 | 2.281e+02 |
| sm90 (#189) | PASS 2.69e-1 | 5.781250e-01 | 2.271e+02 |

**Unit tests** — `tests/test_linear_logp.py`, 27 passed (incl. the SM90 TP tests that exercise the shared backward).

## Files
- `rl_engine/kernels/ops/triton/loss/linear_logp.py` — Triton TP forward + dispatch.
- `rl_engine/kernels/ops/pytorch/loss/linear_logp.py` — shared TP backward helper.
- `rl_engine/kernels/ops/cuda/loss/linear_logp.py` — SM90 TP backward now uses the shared helper.

## Potential Issues: per-forward validation collectives

The exactly-one-owner guarantee comes from `_validate_tp_vocab_partition` and `_validate_global_targets`, which run on every forward and issue collectives on top of the 3 all-reduces the TP reduction itself needs (target logit, running
max, running sum):

- `_validate_tp_vocab_partition` → 2 `all_gather`s (shard ranges + declared global sizes). The vocab partition is fixed at model construction, so this re-proves a static invariant every step.
- `_validate_global_targets` → up to 3 `all_reduce`s (invalid-flag MAX, target MIN, target MAX) plus `.item()` host-device syncs. Data-dependent, but the syncs serialize the stream.

So validation roughly **doubles the collective count per forward**. The messages are tiny (scalar / 2-element), so this is latency-bound, not bandwidth-bound — but on high-rank or high-latency interconnects the extra round-trips (and the host
syncs) can become a **connection/throughput bottleneck** in a tight training loop. This behavior is inherited unchanged from #189 (native + SM90 run the same checks); the Triton path mirrors it for consistency rather than diverging.

## Summary by CodeRabbit

* **Enhancements**
  * Improved log-probability calculations for tensor-parallel and distributed workloads.
  * Added validation for vocabulary and target metadata to identify configuration issues earlier.
  * Improved forward and backward computation across parallel partitions while preserving existing bias handling.
  * Optimized reductions for target logits and normalization values.


Flink-ddd (commented):
Great！Here's a small suggestion for your consideration.

rl_engine/kernels/ops/triton/loss/linear_logp.py
```diff
@@ -154,6 +172,119 @@ def backward(ctx, grad_logp):
...
+        bias_t = bias.contiguous() if bias is not None else hidden_2d
+        vocab_start_index = int(vocab_start_index)
+        global_vocab_size = _validate_tp_vocab_partition(
+            tp_group=tp_group,
+            device=hidden_2d.device,
+            vocab_start_index=vocab_start_index,
+            local_vocab_size=weight.size(0),
+            global_vocab_size=global_vocab_size,
+        )
+        _validate_global_targets(target_1d, global_vocab_size, tp_group)
```
Flink-ddd: Calling _validate_tp_vocab_partition and _validate_global_targets here introduces extra collectives (all_gather, all_reduce) and .item() host syncs on the hot path for every forward step.

Since the TP vocabulary partition is static after initialization, evaluating this on every pass will become a severe latency bottleneck in a distributed training loop. I recommend caching the validation state (e.g., executing it only once per tp_group) or gating these checks behind a validation/debug flag to keep the production forward pass as lightweight as possible.

Flink-ddd (commented):
Great！Here's a small suggestion for your consideration.

Flink-ddd:
please resolve the code conflicts first, then we will merge this PR.

Flink-ddd (approved):

Flink-ddd:
cc @inaniloquentee This is a old PR regarding TP Fused Linear LogP Triton. PTAL when you have available. If there are no issues, we will merge it.

PR #210, opened 2026-07-06
ci: build & test the CUDA extension on the real GPU arch, fail fast i…

### LINKED ISSUE #191: [CI] Support dynamic multi-GPU architecture testing (sm80/86/90/100) to fix NoneType extension load failures
### Problem Description
Currently, running `test_det_gemm.py` in our CI pipeline results in multiple test failures with the following error:

```python
    @staticmethod
    def forward(ctx, a, b):
        ctx.save_for_backward(a, b)
>       return _C.det_gemm_fwd(a, b)
               ^^^^^^^^^^^^^^^
E       AttributeError: 'NoneType' object has no attribute 'det_gemm_fwd'

```

### Root Cause:
This is caused by a CUDA architecture mismatch. The current CI script (ci/run_gpu_ci.sh) hardcodes the compilation target to sm86 (export TORCH_CUDA_ARCH_LIST=8.6). When the cloud provider (RunPod) provisions a GPU with a different architecture, such as an H100 (sm90) or B200 (sm100), PyTorch silently fails to load the compiled .so C++ extension. Consequently, the _C module becomes None, triggering the AttributeError during test execution.

### Proposed Solution
To resolve this and ensure cross-architecture compatibility (including the latest Blackwell series), we need to implement the following pipeline improvements:

### Dynamic GPU Provisioning:
Introduce a TARGET_SM environment variable in ci/run_gpu_ci.sh to dynamically rent the appropriate GPU instance based on the target architecture:
- TARGET_SM=100 → B200 / B100 (sm100)
- TARGET_SM=90 → H100 (sm90)
- TARGET_SM=80 → A100 (sm80)
- TARGET_SM=86 → RTX A4000 / A40 (sm86, default fallback)

### Dynamic Compilation Flags:
Replace the hardcoded TORCH_CUDA_ARCH_LIST with a dynamically injected variable that matches the rented hardware (e.g., 10.0, 9.0, 8.0, or 8.6).

### Fail-Fast Validation:
Add a sanity check (python -c "import torch; import rl_engine") immediately after compilation. If the C++ extension fails to load, the script should exit immediately with a clear architecture mismatch warning, rather than executing the entire test suite and generating misleading logs.

### GitHub Actions Matrix Strategy:
Update the CI workflow YAML to utilize a matrix strategy, triggering parallel test jobs for sm80, sm86, sm90, and sm100. This will provide comprehensive coverage and prevent architecture-specific regressions on future PRs across all supported NVIDIA generations.

Liuzin521:
## Summary

Fixes #191. The reported `_C=None` / "NoneType" failures and the hardcoded-architecture concern are actually **two separate bugs**. This PR fixes both and makes silent degradation impossible.

## Root cause (reproduced)

**Bug A — the real cause of the reported failures is build isolation, not the arch.**
`pyproject.toml`'s `[build-system].requires` does not include `torch`. Under pip's default PEP 517 build isolation, `setup.py` runs in an environment where `import torch` fails, so `get_extensions()` returns `[]` and **`rl_engine._C` is never compiled**. At runtime `rl_engine/kernels/ops/base.py` catches the `ImportError`, logs a warning, and sets `_C = None`.

Because every op has a pure-PyTorch fallback and no test asserts `_C` is present (CUDA tests only `skipif(not torch.cuda.is_available())`), **GPU CI has been passing while the compiled kernels were never built or tested** — it exercises only the native fallbacks. `ci/run_gpu_ci.sh` runs `pip install -e .` with isolation on,so this is the current state.

```console
$ pip install -e .            # isolation on (default)
$ ls rl_engine/_C*.so         # -> nothing built
$ python -c "from rl_engine import _C"
ImportError: cannot import name '_C' from 'rl_engine'
```

**Bug B — the hardcoded arch is a real but latent bug.**
Once the extension does build, `TORCH_CUDA_ARCH_LIST=8.6` compiles SASS only for sm_86 with no PTX. On a non-sm86 GPU, `import _C` **succeeds** (dlopen does not check GPU arch) but the first kernel launch raises `cudaErrorNoKernelImageForDevice`. That is not an `ImportError`, so `base.py` cannot catch it.

```console
$ TORCH_CUDA_ARCH_LIST=8.6 pip install --no-build-isolation -e .   # sm_86 .so
# launched on a newer GPU:
RuntimeError: CUDA error: no kernel image is available for execution on the device
```

## Changes

- **`ci/run_gpu_ci.sh`**
  - Install with `--no-build-isolation` so torch is visible to the build (fixes Bug A).
  - Detect the pod GPU's compute capability (`nvidia-smi --query-gpu=compute_cap`, with a torch fallback) instead of hardcoding `8.6`, and build with `+PTX` for forward compatibility (fixes Bug B).
  - An explicit `TARGET_SM` (for a future arch matrix) is normalized and **asserted against the actual pod GPU**, so a cross-architecture resource fallback fails fast instead of silently building mismatched SASS.
  - Log the torch version before install (the CI image ships torch 2.4.0 while the project requires `>=2.4.1`).
- **`scripts/ci_smoke.py`** (new) — fail-fast check run before pytest: imports `_C` and launches `fused_logp` (the unconditionally-registered op) followed by `torch.cuda.synchronize()`, so both failure modes above fail loudly with an actionable message.
- **`tests/test_extension_smoke.py`** (new) — gated regression test: on a GPU host `_C` must be present and launchable; on CPU it skips (the CPU CI job legitimately has no `_C`).
- **`docs/getting_started/installation.md`** — document `--no-build-isolation` for source builds; CPU-only installs remain supported on the pure-PyTorch backends.
- **`.github/workflows/gpu-ci.yml`** — matrix-ready scaffold (sm80/86/90/100) documenting how `GPU_ID` / `TARGET_SM` map to the script.

`setup.py`, `pyproject.toml`, and `rl_engine/kernels/ops/base.py` are intentionally unchanged, preserving graceful runtime fallback for legitimate CPU/no-GPU installs.

## Verification

Reproduced on an RTX PRO 6000 Blackwell (sm_120) host, torch 2.11+cu128, nvcc 12.8:

| Scenario | `scripts/ci_smoke.py` | `pytest` |
| --- | --- | --- |
| Default isolation → no `.so` (Bug A) | exit 1 (missing) | fail |
| sm_86 build on sm_120 (Bug B) | exit 1 (launch error) | fail |
| Auto-detect → sm_120 + PTX (fixed) | exit 0 | pass |

`cuobjdump` confirms the fixed build embeds both `sm_120` SASS and `sm_120` PTX.

## Note on this PR's own GPU CI

`gpu-ci.yml` uses `pull_request_target` and checks out the orchestrator
(`ci/run_gpu_ci.sh`) from the **base** branch for security. So this PR's own GPU CI
would run the *current, unfixed* orchestrator against this branch's code, which means:

- the `run_gpu_ci.sh` changes here are not exercised until they land on `main`, and
- `tests/test_extension_smoke.py` will correctly go **red** under the old orchestrator
  (it detects that the extension was never built) — which is exactly the bug this PR fixes.

The fix is therefore validated locally (see Verification above). To see it green in live
CI, the new `run_gpu_ci.sh` needs to run from a base-repo branch, or simply after merge.

## Follow-up

Enabling the multi-arch matrix (sm80/90/100) needs RunPod GPU IDs the maintainer provisions. The script is already parametrized (`GPU_ID` + `TARGET_SM`) and asserts the arch, so it is ready to wire up; per-arch jobs must not fall back across compute capabilities.

## Summary by CodeRabbit

* **New Features**
  * Added a fail-fast CUDA smoke check to verify the compiled native extension loads and that `fused_logp` runs correctly on the active GPU.
  * Added a GPU-only regression test that hard-fails if the extension is missing or kernel execution fails (opt-in via an environment setting).
* **Bug Fixes**
  * Enhanced GPU CI to select GPUs and target compute capability from environment settings, validate requested targets against the provisioned GPU, and exit early on mismatches.
* **Documentation**
  * Expanded “From Source” installation guidance for CUDA/ROCm, including PyTorch-first steps, build isolation requirements, optional architecture pinning, and extension verification.

### DIFF
--- a/.github/workflows/gpu-ci.yml
+++ b/.github/workflows/gpu-ci.yml
@@ -48,9 +48,24 @@ jobs:
           chmod 600 ~/.ssh/id_ed25519
           ssh-keygen -y -f ~/.ssh/id_ed25519 > /dev/null && echo "key OK" || echo "key BROKEN"
 
+      # Follow-up (#191): to test multiple architectures, add a matrix and pass
+      # GPU_ID + TARGET_SM through to the script (run_gpu_ci.sh reads both; it
+      # normalizes TARGET_SM and asserts the provisioned pod actually matches it):
+      #   strategy:
+      #     matrix:
+      #       include:
+      #         - { gpu_id: "NVIDIA RTX A4000",      target_sm: "8.6" }   # Ampere
+      #         - { gpu_id: "NVIDIA A100 80GB PCIe", target_sm: "8.0" }
+      #         - { gpu_id: "NVIDIA H100 PCIe",      target_sm: "9.0" }   # + KERNEL_ALIGN_FORCE_SM90=1 to exercise Hopper kernels
+      #         - { gpu_id: "NVIDIA B200",           target_sm: "10.0" }
+      # Per-arch jobs must NOT fall back to a different-capability GPU: the script
+      # fails fast when the pod arch != requested TARGET_SM, so keep fallback within
+      # the same compute capability (or unset it for these jobs).
       - name: Run GPU tests on RunPod
         env:
           RUNPOD_API_KEY: ${{ secrets.RUNPOD_API_KEY }}
           PR_REPO_URL: ${{ github.event.pull_request.head.repo.clone_url }}
           PR_SHA: ${{ github.event.pull_request.head.sha }}
+          # GPU_ID: ${{ matrix.gpu_id }}
+          # TARGET_SM: ${{ matrix.target_sm }}
         run: bash ci/run_gpu_ci.sh
--- a/.github/workflows/gpu-ci.yml
+++ b/.github/workflows/gpu-ci.yml
@@ -49,14 +49,15 @@ jobs:
           ssh-keygen -y -f ~/.ssh/id_ed25519 > /dev/null && echo "key OK" || echo "key BROKEN"
 
       # Follow-up (#191): to test multiple architectures, add a matrix and pass
-      # GPU_ID + TARGET_SM through to the script (run_gpu_ci.sh reads both; it
-      # normalizes TARGET_SM and asserts the provisioned pod actually matches it):
+      # GPU_ID + TARGET_SM (+ KERNEL_ALIGN_FORCE_SM90 for Hopper) through to the script.
+      # run_gpu_ci.sh reads all three, normalizes TARGET_SM, asserts the pod matches it,
+      # and forwards KERNEL_ALIGN_FORCE_SM90 into the remote build:
       #   strategy:
       #     matrix:
       #       include:
       #         - { gpu_id: "NVIDIA RTX A4000",      target_sm: "8.6" }   # Ampere
       #         - { gpu_id: "NVIDIA A100 80GB PCIe", target_sm: "8.0" }
-      #         - { gpu_id: "NVIDIA H100 PCIe",      target_sm: "9.0" }   # + KERNEL_ALIGN_FORCE_SM90=1 to exercise Hopper kernels
+      #         - { gpu_id: "NVIDIA H100 PCIe",      target_sm: "9.0", force_sm90: "1" }  # build Hopper TMA/WGMMA kernels
       #         - { gpu_id: "NVIDIA B200",           target_sm: "10.0" }
       # Per-arch jobs must NOT fall back to a different-capability GPU: the script
       # fails fast when the pod arch != requested TARGET_SM, so keep fallback within
@@ -68,4 +69,5 @@ jobs:
           PR_SHA: ${{ github.event.pull_request.head.sha }}
           # GPU_ID: ${{ matrix.gpu_id }}
           # TARGET_SM: ${{ matrix.target_sm }}
+          # KERNEL_ALIGN_FORCE_SM90: ${{ matrix.force_sm90 }}
         run: bash ci/run_gpu_ci.sh
--- a/ci/run_gpu_ci.sh
+++ b/ci/run_gpu_ci.sh
@@ -9,15 +9,11 @@ PRIMARY_GPU_COUNT="${PRIMARY_GPU_COUNT:-${GPU_COUNT:-2}}"
 FALLBACK_GPU_ID="${FALLBACK_GPU_ID:-NVIDIA A40}"
 FALLBACK_GPU_COUNT="${FALLBACK_GPU_COUNT:-1}"
 
-# Optional explicit compile target (e.g. "9.0" or "90"). When set it is treated as
-# an ASSERTION: the provisioned pod must actually be this arch (checked on the pod in
-# the remote build block), so a cross-architecture resource fallback fails fast rather
-# than silently compiling+launching mismatched SASS.
+# Optional arch override; asserted against the pod's real cap in the remote build so a
+# cross-arch resource fallback cannot build mismatched SASS.
 TARGET_SM="${TARGET_SM:-}"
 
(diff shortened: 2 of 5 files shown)


Flink-ddd:
cc @KJLdefeated @a-kaa @inaniloquentee PTAL, this PR is very import. When testing operator determinism, we also need to ensure that all operators pass GPU testing and there are no regression issues. This PR can support multi-GPU architectures, such as SM 80, 86, 90, 100, etc., instead of just launching the initial GPU resource test for SM8.6.

.github/workflows/gpu-ci.yml
```diff
@@ -48,9 +48,24 @@ jobs:
...
+      # GPU_ID + TARGET_SM through to the script (run_gpu_ci.sh reads both; it
+      # normalizes TARGET_SM and asserts the provisioned pod actually matches it):
+      #   strategy:
+      #     matrix:
+      #       include:
+      #         - { gpu_id: "NVIDIA RTX A4000",      target_sm: "8.6" }   # Ampere
+      #         - { gpu_id: "NVIDIA A100 80GB PCIe", target_sm: "8.0" }
+      #         - { gpu_id: "NVIDIA H100 PCIe",      target_sm: "9.0" }   # + KERNEL_ALIGN_FORCE_SM90=1 to exercise Hopper kernels
+      #         - { gpu_id: "NVIDIA B200",           target_sm: "10.0" }
+      # Per-arch jobs must NOT fall back to a different-capability GPU: the script
```
KJLdefeated: Pls fix the linting pre-commit failed

KJLdefeated (changes requested):
Overall looks good and the smoke test is exactly the right guard, but please fix the current lint failure before merge. Also please make the CI torch version/build-order deterministic, since the image ships torch 2.4.0 while the package requires >=2.4.1. For the future SM90 matrix, make sure KERNEL_ALIGN_FORCE_SM90 is actually forwarded into the remote build; otherwise H100 jobs won’t exercise the Hopper kernels.

Liuzin521:
@Flink-ddd Heads-up on the red `linting` check: it's failing on files this PR doesn't touch —
`rl_engine/kernels/gtest/*`, `scripts/check_operator.py`, `tests/test_op_checks.py`
(trailing-whitespace / black / isort). Those landed via recent merges  and
since CI runs `pre-commit --all-files` on the merge with main, those repo-wide violations
surface on every PR. `main` itself is currently red on lint for the same reason.

This PR's own files are lint-clean (`pre-commit run --all-files` passes on them).

KJLdefeated (approved):
The correctness issues I raised look resolved, so I’m not blocking on that anymore. My only remaining concern is readability: the current CI script now contains several long explanatory comment blocks, many of which duplicate the PR description. Since this file is an operational script, I’d prefer keeping comments focused on durable invariants and removing historical/debugging context that belongs in the PR body or issue discussion.

Could you trim the comments down to the essential “why this must stay this way” notes before merge?

Flink-ddd:
@Liuzin521 Could you please resolve this CI check issue? Aside from this problem, there's nothing else.

Flink-ddd (approved):
LGTM, let's merge this PR first and run our kernel PR. Thank you for work!

Liuzin521:
> @Liuzin521 Could you please resolve this CI check issue? Aside from this problem, there's nothing else.

@Flink-ddd opened the lint cleanup as a separate PR to keep this one focused: #213.
It's `pre-commit run --all-files` auto-fixes only (black / isort / trailing-whitespace)
on the files `main` picked up — no behavior change. Once #213 merges, the `linting`
check here goes green too.

Issue #45, opened 2026-05-31
[FEAT][kernels]: implement fused Gumbel-Softmax sampler for differentiable rollouts


Flink-ddd:
### Context:
Standard autoregressive sampling like greedy or temperature sampling in vLLM breaks the computational graph, requiring REINFORCE-style gradient estimators like PPO. For future alignment algorithms that may rely on reparameterization gradients through discrete tokens, we need a high-performance, differentiable Gumbel-Softmax sampling operator.

## Tasks:
1. Implement a Triton kernel for the Gumbel-Softmax trick in rl_engine/kernels/ops/triton/sampling.py.
2. The kernel should support hard=True returns one-hot vectors on forward, continuous gradients on backward and temperature annealing.
3. Ensure it efficiently handles the [Batch, Sequence, Vocab_size] tensor shape without materializing excessive intermediate random noise arrays in global memory.

JLiu4Coding:
Hi Vensen, I am eagerly working on this. Can this be assigned to me?

Flink-ddd:
for sure, assigned, Thank you.

JLiu4Coding:
Hi @Flink-ddd , I have created a draft PR #215  addressing this issue. The correctness has successfully passed and our triton implementation has outperformed the native pytorch implementation in terms of both speed and saving memory. Willing to further revise as needed.

PR #215, opened 2026-07-10
[FEAT][kernels] Fused Gumbel-Softmax sampler — Triton forward, ST hard path, fused backward

JLiu4Coding:
## Summary

This PR addressed the issue #45 by adding a fused **Gumbel-Softmax sampler operator** for differentiable rollout and sampling workloads.

### Implemented Backends

- PyTorch reference implementation
- Triton implementation

### Supported Features

- Soft Gumbel-Softmax output with `hard=False`
- Straight-through one-hot output with `hard=True`
- Optional caller-supplied Gumbel noise for deterministic testing
- Backend-internal Gumbel noise generation
- Backward support for `logits`
- Registry dispatch through `gumbel_softmax`
- Operator documentation, tests, and benchmarks

The Triton backend also includes a specialized no-grad `hard=True` fast path that skips the softmax computation and writes the one-hot output directly.

---

## Correctness

Tested with the following environment:

| Component | Version |
|---|---|
| GPU | NVIDIA A40 |
| Platform | RunPod |
| CUDA | 12.4 |
| PyTorch | 2.4.1+cu124 |
| Triton | 3.0.0 |

### Test Command

```bash
python -m pytest tests/test_gumbel_softmax.py -q
```

### Test Result

```text
11 passed in 3.90s
```

---

## Benchmark

The benchmark uses 10 measured iterations after 3 warm-up iterations.

By default, caller-supplied Gumbel noise is used to provide a deterministic, apples-to-apples comparison between the PyTorch and Triton implementations.

#### Command

```bash
PYTHONPATH=. python benchmarks/benchmark_gumbel_softmax.py \
  --iters 10 \
  --warmup 3
```

#### Results

| Shape `(B × S × V)` | Dtype | Native Fwd | Triton Fwd | Fwd Speedup | Native Fwd+Bwd | Triton Fwd+Bwd | Fwd+Bwd Speedup | Native Fwd Memory | Triton Fwd Memory |
|---|---|---:|---:|---:|---:|---:|---:|---:|---:|
| `1 × 512 × 32000` | `float16` | 0.632 ms | 0.488 ms | **1.29×** | 1.274 ms | 0.766 ms | **1.66×** | 95 MB | 31 MB |
| `4 × 512 × 32000` | `float16` | 2.450 ms | 0.900 ms | **2.68×** | 4.971 ms | 2.961 ms | **1.68×** | 377 MB | 125 MB |
| `4 × 1024 × 50257` | `float16` | 8.157 ms | 3.857 ms | **2.11×** | 16.170 ms | 10.416 ms | **1.55×** | 1179 MB | 393 MB |

#### Command

```bash
PYTHONPATH=. python benchmarks/benchmark_gumbel_softmax.py \
  --iters 10 \
  --warmup 3 \
  --hard
```

#### Results

| Shape `(B × S × V)` | Dtype | Native Fwd | Triton Fwd | Fwd Speedup | Native Fwd+Bwd | Triton Fwd+Bwd | Fwd+Bwd Speedup | Native Fwd Memory | Triton Fwd Memory |
|---|---|---:|---:|---:|---:|---:|---:|---:|---:|
| `1 × 512 × 32000` | `float16` | 1.117 ms | 0.238 ms | **4.70×** | 1.761 ms | 1.060 ms | **1.66×** | 159 MB | 31 MB |
| `4 × 512 × 32000` | `float16` | 4.298 ms | 0.900 ms | **4.78×** | 6.816 ms | 3.540 ms | **1.93×** | 629 MB | 125 MB |
| `4 × 1024 × 50257` | `float16` | 13.869 ms | 3.067 ms | **4.52×** | 21.899 ms | 20.320 ms | **1.08×** | 1964 MB | 393 MB |

---

## Key Results

- Soft-mode forward execution is up to **2.68× faster**.
- Hard-mode forward execution is up to **4.78× faster**.
- Forward-and-backward execution is up to **1.93× faster**.
- Triton forward memory usage is approximately **3–5× lower** across the tested configurations.
- The no-grad hard-mode fast path avoids computing and materializing soft probabilities.

---

## Implementation Notes

- The PyTorch backend serves as the correctness reference.
- The Triton backend stores saved soft probabilities using the same dtype as `logits` to reduce memory usage.
- No-grad `hard=True` calls use a dedicated fast path that skips softmax and writes only the one-hot output.
- Backward propagation for straight-through hard samples follows the softmax gradient path.
- The benchmark uses caller-supplied Gumbel noise by default for deterministic backend comparisons.
- Use `--internal-gumbels` to benchmark backend-internal Gumbel noise generation.

## Summary by CodeRabbit

* **New Features**
  * Added `gumbel_softmax` with both soft probabilities and straight-through hard (one-hot) sampling, including temperature control and deterministic noise support.
  * Introduced PyTorch and Triton GPU implementations with automatic backend dispatch and gradient-aware backward.
* **Documentation**
  * Added operator documentation and linked it from the operators index.
* **Tests**
  * Expanded unit tests for correctness, shapes/dtypes, gradients, validation, and Triton GPU paths.
* **Benchmarks**
  * Added a benchmark comparing native vs GPU performance and peak memory usage.
* **Chores**
  * Minor CI smoke/test robustness and header/lint cleanups.


Flink-ddd:
cc @KJLdefeated @inaniloquentee @EthanZero2Hero PTAL

JLiu4Coding:
I pushed the reviewer fixes and pre-commit formatting updates.

Local validation:
- black passed
- isort passed
- flake8 passed
- `python -m pytest tests/test_gumbel_softmax.py -q` passed locally with CUDA tests skipped

The latest CI run is awaiting maintainer approval. Could you please approve/re-run the workflow when convenient?

Flink-ddd:
cc @KJLdefeated @inaniloquentee @maxiaosong1124 PTAL

rl_engine/kernels/registry.py
```diff
@@ -155,20 +114,16 @@ def __init__(self):
...
-                    OpBackend.CUDA_DETERMINISTIC_LOGP,
-                    OpBackend.PYTORCH_NATIVE,
-                ],
                 "attn": [OpBackend.FLASH_ATTN, OpBackend.TRITON_GENERIC, OpBackend.PYTORCH_ATTN],
                 "attention": [OpBackend.PYTORCH_NATIVE_ATTENTION],
                 "kv_cache_attention": [OpBackend.PYTORCH_NATIVE_KV_CACHE_ATTN],
                 "grpo_loss": [OpBackend.TRITON_GRPO_LOSS, OpBackend.PYTORCH_GRPO_LOSS],
                 "linear_logp": [OpBackend.TRITON_LINEAR_LOGP, OpBackend.PYTORCH_LINEAR_LOGP],
                 "ratio_kl": [OpBackend.TRITON_RATIO_KL, OpBackend.PYTORCH_RATIO_KL],
+                "gumbel_softmax": [
```
inaniloquentee, line 218: One concern here: on GPU the registry will pick the Triton backend by default, but this implementation hard-fails when vocab_size > 131072. That means larger LLM vocabularies, e.g. 151936, will crash instead of falling back to the Native backend.

Could we either add a shape-aware fallback for this case, or avoid routing to Triton by default when the vocab is above the supported limit?

tests/test_gumbel_softmax.py
```diff
@@ -0,0 +1,167 @@
...
+        op(logits, gumbels=gumbels[:, :-1])
+    with pytest.raises(ValueError, match="at least 2 dimensions"):
+        op(torch.randn(8))
+
+
+def test_registry_dispatch_matches_native():
+    from rl_engine.kernels.registry import kernel_registry
+    from rl_engine.platforms.device import device_ctx
+
+    device = device_ctx.device if device_ctx.device_type == "cuda" else "cpu"
```
inaniloquentee: Small but important test issue: on ROCm, device_ctx.device_type is "rocm", so this test puts the inputs on CPU. But kernel_registry.get_op("gumbel_softmax") will still select the ROCm/Triton backend, which then receives a CPU tensor and fails.

Can we make the test use the actual GPU device when device_ctx.is_rocm is true as well?

rl_engine/kernels/ops/triton/sampling/gumbel_softmax.py
```diff
@@ -0,0 +1,210 @@
...
+        grad_2d = grad_output.contiguous().view_as(y_soft)
+        grad_logits = torch._softmax_backward_data(
+            grad_2d,
+            y_soft,
+            -1,
+            ctx.logits_dtype,
+        )
+        if ctx.tau != 1.0:
+            grad_logits = grad_logits / ctx.tau
+        return grad_logits.view(ctx.logits_shape).to(ctx.logits_dtype), None, None, None, None
```
inaniloquentee, line 322: The Native and Triton backends currently differ if the supplied gumbels tensor requires grad: Native propagates gradients through gumbels, while Triton always returns None for that input.

If gumbels is intended to be treated as fixed noise, that is totally reasonable, but then we should probably detach it in the Native path too or document that contract explicitly.

inaniloquentee (changes requested):

rl_engine/kernels/registry.py
JLiu4Coding: Thanks for catching this. This is indeed an delicate yet important issue. Fixed by adding a shape-aware fallback inside `TritonGumbelSoftmaxOp`: when `triton.next_power_of_2(V)` exceeds the Triton block limit (`131072`), the wrapper now falls back to `NativeGumbelSoftmaxOp` instead of raising. This keeps Triton as the default for supported vocab sizes while avoiding crashes for larger LLM vocabularies.

tests/test_gumbel_softmax.py
JLiu4Coding: Fixed. The registry dispatch test now treats ROCm as a GPU path too, so it uses `device_ctx.device` when either CUDA is active or `device_ctx.is_rocm` is true:

```python
device = device_ctx.device if device_ctx.device_type == "cuda" or device_ctx.is_rocm else "cpu"

File: rl_engine/kernels/ops/triton/sampling/gumbel_softmax.py (hunk 3 above)
Comment (JLiu4Coding) [reply rl_engine/kernels/ops/triton/sampling/gumbel_softmax.py:322]: Thanks, fixed. I made the contract consistent by treating supplied `gumbels` as fixed sampling noise in both backends. The Native path now detaches supplied `gumbels`, matching the Triton backward behavior, and the docs explicitly state that gradients are not propagated to `gumbels`.

**Comment by JLiu4Coding** [comment]
Hi @inaniloquentee , addressed the latest review comments in the new commit:

- Added large-vocab fallback from Triton to Native when the Triton block limit is exceeded.
- Fixed the registry dispatch test to use the GPU device on ROCm as well as CUDA.
- Made supplied `gumbels` fixed-noise semantics consistent across Native and Triton, and documented that gradients are not propagated to `gumbels`.

Local validation:
- `python -m pytest tests/test_gumbel_softmax.py -q`
- `black --check`
- `isort --check-only`
- `flake8`

All passed locally; CUDA-specific tests are skipped on this machine and should run in CI/GPU environment.

**Comment by JLiu4Coding** [comment]
Oops! There're some CI issues here. Let me have a look. I will follow up with a new PR asap

**Comment by JLiu4Coding** [comment]
Hi @inaniloquentee , would you please review again at your convenience? I believe your previous comments have been addressed.

File: rl_engine/kernels/ops/triton/sampling/gumbel_softmax.py [hunk 4]
```diff
@@ -0,0 +1,215 @@
...
+
+
+def _launch_config(vocab_size: int) -> tuple[int, int]:
+    block_v = triton.next_power_of_2(vocab_size)
+    if block_v > _MAX_BLOCK_V:
+        raise ValueError(
+            f"vocab size {vocab_size} exceeds TritonGumbelSoftmaxOp limit {_MAX_BLOCK_V}"
+        )
+    num_warps = 32 if block_v >= 65536 else 16 if block_v >= 32768 else 8
+    return block_v, num_warps
```
KJLdefeated, line 26: The triton kernel did not cover our target model (Qwen3). Qwen3's vocab is 151936. I would suggest modification to fit Qwen3's config.

rl_engine/kernels/ops/triton/sampling/gumbel_softmax.py
```diff
@@ -0,0 +1,215 @@
...
+
+    tl.store(y_soft_ptr + row_off + cols, y, mask=mask)
+    if HARD:
+        # Tie-break to the first maximum so each row is exactly one-hot.
+        is_max = z == z_max
+        first_max = tl.min(tl.where(is_max, cols, BLOCK_V), axis=0)
+        out = tl.where(cols == first_max, 1.0, 0.0)
+        tl.store(out_ptr + row_off + cols, out, mask=mask)
+    else:
+        tl.store(out_ptr + row_off + cols, y, mask=mask)
```
KJLdefeated, line 93: When HARD==False, we would do second write on tl.store(y_soft_ptr + row_off + cols, y, mask=mask)?
Pls correct me if I am wrong.

rl_engine/kernels/ops/triton/sampling/gumbel_softmax.py
```diff
@@ -0,0 +1,215 @@
...
+    row = tl.program_id(0)
+    cols = tl.arange(0, BLOCK_V)
+    mask = cols < V
+    row_off = row.to(tl.int64) * V
+
+    logits = tl.load(logits_ptr + row_off + cols, mask=mask, other=-float("inf")).to(tl.float32)
+    if HAS_GUMBELS:
+        gumbels = tl.load(gumbels_ptr + row_off + cols, mask=mask, other=0.0).to(tl.float32)
+    else:
+        u = tl.rand(seed, row_off + cols)
```
KJLdefeated, line 75: `tl.rand(seed, row_off + cols)` with an int64 offset. row_off is explicitly cast to int64, so the offset expression is int64, but Philox in Triton is int32-based. For N*V > 2^31 this either wraps (silent noise-stream collisions across rows) or fails to compile. Worth a test at large N.

KJLdefeated (changes requested):

rl_engine/kernels/ops/triton/sampling/gumbel_softmax.py
JLiu4Coding: Thanks for pointing this out. `_MAX_BLOCK_V = 131072` was chosen because the current kernel uses one Triton program per row, with `BLOCK_V = next_power_of_2(V)`. This covers common vocab sizes up to around 128k.

For Qwen3, `V = 151936`, so `next_power_of_2(V) = 262144`. The current fallback avoids crashing, but I agree it means the Triton path does not really cover the target Qwen3 case.

I see two possible directions:
1. Increase `_MAX_BLOCK_V` to `262144`, which is simpler, but may create a very large single-program reduction and could hurt compile stability/performance.
2. Implement a chunked/tiled Triton path for large vocab sizes, which should be more scalable but is more complex.

Do you have any suggestion on the direction or any other comment?

rl_engine/kernels/ops/triton/sampling/gumbel_softmax.py
JLiu4Coding: Yes, you are right. When hard=False, out is the same tensor as y_soft, so the second tl.store was redundant.

Fixed by adding a STORE_OUT constexpr flag: soft mode stores once, while hard mode still stores both y_soft for backward and the one-hot output. I also applied the same fix to the chunked Triton path.

rl_engine/kernels/ops/triton/sampling/gumbel_softmax.py
JLiu4Coding: I went with the chunked/tiled Triton path instead of simply increasing _MAX_BLOCK_V to 262144.

The updated implementation keeps the existing one-block kernel for supported vocab sizes, and adds a separate large-vocab Triton path for cases like Qwen3 (V=151936). The chunked path splits the vocab dimension into chunks, computes chunk-local max/sum statistics, reduces them into row-global softmax statistics, and then writes the final output. So it still computes one correct global softmax over the full vocab, not independent per-chunk softmaxes.

I also added a Qwen3-vocab GPU smoke test with shape (1, 151936), plus chunked-path tests for soft forward, hard forward, hard no-grad, and backward correctness against the Native reference.

Validation:
- RunPod A40: python -m pytest tests/test_gumbel_softmax.py -q -s -> 19 passed
- GPU CI A4000/A40: passed
- GPU CI H100: passed

rl_engine/kernels/ops/triton/sampling/gumbel_softmax.py
JLiu4Coding: I addressed this by avoiding the flattened tl.rand(seed, row_off + cols) path for large/unsafe shapes.

The wrapper now checks whether the tensor needs precomputed Gumbel noise. If the vocab requires the chunked large-vocab path, or if logits.numel() exceeds the safe RNG offset range, it precomputes Gumbel noise with PyTorch and passes that tensor into Triton. In that case the Triton kernel only does tl.load(gumbels_ptr + ...), so it no longer depends on a potentially overflowing Philox offset.

For smaller shapes, the existing in-kernel tl.rand path is kept.

I also added tests for the precompute guard and validated the full Gumbel-Softmax test file on RunPod A40:
19 passed.
Both GPU CI jobs also passed.

JLiu4Coding:
Hi @KJLdefeated, thanks for your comments. Those make a lot of sense. I have addressed. Could you please have a review again at your convenience?

Flink-ddd:
please resolve the code conflicts first, Thank you.

JLiu4Coding:
Hi @inaniloquentee and @KJLdefeated, the branch is now updated with the latest main, conflicts are resolved, and both the A4000 and H100 GPU CI jobs pass. I have addressed the previous review points, including Qwen3 large-vocabulary chunking, safe Gumbel RNG handling, fixed-noise gradient semantics, ROCm test device selection, and the redundant soft-mode stores. Could you please re-review the latest head (e6b55d7)  when convenient? Thank you!

Flink-ddd:
cc @KJLdefeated @inaniloquentee PTAL again, Thank you.

Flink-ddd (approved):
LGTM, Thank you for work!

JLiu4Coding:
This failure looks like RunPod capacity rather than a code/test failure. The A4000 job could not provision either the requested A4000 pod or the A40 fallback:

`failed to create pod: graphql error: This machine does not have the resources to deploy your pod`

The H100 GPU job passed. Could we rerun the failed A4000 job when RunPod capacity is available?

Flink-ddd:
> This failure looks like RunPod capacity rather than a code/test failure. The A4000 job could not provision either the requested A4000 pod or the A40 fallback:
> 
> `failed to create pod: graphql error: This machine does not have the resources to deploy your pod`
> 
> The H100 GPU job passed. Could we rerun the failed A4000 job when RunPod capacity is available?

No need to rerun the A4000 job. We are going to remove this CI check since we're now targeting the Hopper architecture. However, we'd like to keep this PR open. If possible, could you adapt it for the Hopper architecture? Thank you.

JLiu4Coding:
> > This failure looks like RunPod capacity rather than a code/test failure. The A4000 job could not provision either the requested A4000 pod or the A40 fallback:
> > `failed to create pod: graphql error: This machine does not have the resources to deploy your pod`
> > The H100 GPU job passed. Could we rerun the failed A4000 job when RunPod capacity is available?
> 
> No need to rerun the A4000 job. We are going to remove this CI check since we're now targeting the Hopper architecture. However, we'd like to keep this PR open. If possible, could you adapt it for the Hopper architecture? Thank you.

Thanks for the clarification. I validated the current implementation on Hopper/H100.

Environment:
- GPU: NVIDIA H100 80GB HBM3
- Compute capability: SM90 / 9.0
- CUDA: 12.4
- PyTorch: 2.4.1+cu124
- Triton: 3.0.0

Validation:
- `python -m pytest tests/test_gumbel_softmax.py -q -s`
- Result: `19 passed in 6.27s`

I also ran a direct Qwen3-vocab smoke test with shape `(1, 151936)`; output was finite and the row sum was `1.0`.

The implementation is not SM86-specific and the chunked Triton path works on SM90. I can also update the PR description to make the Hopper/H100 validation the primary result and de-emphasize the older A40/A4000 validation.

PR #224, opened 2026-07-14
  [WS1]test: relax native matmul batch-gradient comparison

frank-2077:
## Summary
  #225 
  - Replace the strict bitwise comparison in `TestNativeMatmulOpBatchInvariance.test_batch_grad_invariance` with an explicit numerical tolerance.
  - Document why splitting a batch changes GEMM's M dimension and can change the floating-point reduction order even under single-threaded CPU execution.

 ## Problem

  The test compared the input gradients from a full-batch GEMM and per-row GEMMs using `torch.equal`.

  These executions are mathematically equivalent, but native PyTorch GEMM does not guarantee bitwise-identical floating-point results across different matrix shapes or backend dispatches. In GPU CI, this CPU-tensor test ran on different pod hosts and produced normal rounding drift of approximately `1e-4`, blocking unrelated kernel PRs.

## Summary by CodeRabbit

* **Tests**
  * Improved gradient validation for batched matrix multiplication by allowing small numerical differences caused by operation ordering.
  * Added explanatory coverage for expected floating-point variation when batch rows are split.
* **Chores**
  * Updated GPU CI job matrix to adjust the set of targeted GPU configurations.

### DIFF
--- a/.github/workflows/gpu-ci.yml
+++ b/.github/workflows/gpu-ci.yml
@@ -31,7 +31,6 @@ jobs:
       matrix:
         include:
           - { gpu_id: "NVIDIA RTX A4000",      target_sm: "8.6" }   # SM86
-          - { gpu_id: "NVIDIA A100 80GB PCIe", target_sm: "8.0" }   # SM80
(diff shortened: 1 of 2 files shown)


Flink-ddd (approved):
LGTM, Thank you for fix this issue.

PR #233, opened 2026-07-19
feat(kernels): Triton FlashAttention LSE export + varlen packing

Billy1900:
## Summary

First step of a stack toward "Fused FlashAttention with causal mask, varlen
packing, and exported attention LSE" (see discussion in the repo). Rather than
attempting SM90 WGMMA+TMA + SM80 + ROCm MFMA + Triton all in one PR, this
lands the Triton piece first: it's the only backend where a real kernel
already existed to extend (`rl_engine/kernels/ops/triton/triton_attn.py`), and
it becomes the cross-platform semantic baseline the SM90/SM80/ROCm kernels
get checked against later, rather than against each other.

## What this adds

- **Attention-domain LSE export**: `return_lse=True` on the existing dense
  `triton_flash_attention`. No new kernel math — `M` (running max) and `L`
  (running sum-exp) were already computed online for the backward pass, just
  never returned. LSE is marked non-differentiable (`ctx.mark_non_differentiable`),
  matching `flash_attn`'s external `softmax_lse` contract. Default calls
  (`return_lse=False`) are unchanged — no breaking change to the (currently
  uncalled) public API.
- **`triton_flash_attention_varlen`**: a new packed variable-length path over
  `cu_seqlens`-style `[total_tokens, H, D]` tensors (same convention as
  `flash_attn_varlen_func` and this repo's `pack` op, #182), with forward +
  backward kernels. Causal masking uses the `Skv - Sq` anchor from
  `NativeAttentionOp` (docs/operators/attention.md) so prefill (`Sq == Skv`)
  and decode (`Sq < Skv`) share one formula.

**Explicitly out of scope for this PR**: SM90 WGMMA+TMA, SM80 `mma.sync`, and
ROCm MFMA kernels, and GQA support (the pre-existing dense Triton kernel
didn't support GQA either — not a regression). Not wired into
`KernelRegistry` yet.

## Correctness

Both paths are checked against an **independent** fp32 masked-softmax +
logsumexp reference (not against each other, so a shared bug in the
online-softmax loop can't hide). Run end-to-end on real H100 SXM5 hardware:

- Dense: `out` max-abs-diff ≈ `1.4e-3`, `lse` ≈ `1e-6` vs. reference (fp16
  inputs, fp32 accumulation).
- Varlen: forward + backward (`dq`/`dk`/`dv`) checked across causal/non-causal,
  decode-style (`Sq < Skv`, varying independently per sequence in the batch),
[...]

PR #237, opened 2026-07-20
feat(kernels): SM90 TMA + mma.sync FlashAttention, forward + backward (causal, varlen, LSE)

Billy1900:
## Summary
- Adds the CUDA SM90 (Hopper) FlashAttention kernel, forward **and backward**: causal masking, packed `cu_seqlens` variable-length input, attention-domain LSE export, and `dQ`/`dK`/`dV`, matching the semantics of `triton_flash_attention_varlen` (the cross-platform baseline from #233).
- Compute path is TMA (global->shared loads) + PTX `mma.sync.aligned.m16n8k16` -- the same combination every other "SM90" kernel in this repo already uses, not literal `wgmma.mma_async`.
- Backward is recompute-based (grid over KV-tiles, looping Q-tiles), reusing the forward's saved `lse` directly rather than separate M/L state. `dQ` is atomic-accumulated directly in bf16 (native SM90 `atom.add.noftz.bf16`, not an emulated CAS loop -- verified against `cuda_bf16.hpp`); `dK`/`dV` are written once per KV-tile CTA, no atomics needed there.
- The Python wrapper is now a real `torch.autograd.Function` -- previously `flash_attention_sm90_varlen` called the raw op directly with no autograd wrapping at all, so gradients silently did not flow to `q`/`k`/`v`. Dispatch requires both the forward and backward native symbols before routing to the hardware path; missing either falls back to `triton_flash_attention_varlen` entirely (never mixes an SM90 forward with a Triton backward).
- bf16, `head_dim=128`, contiguous, no GQA. Falls back to `triton_flash_attention_varlen` for anything outside that support surface rather than erroring.
- SM80 `mma.sync` fallback, ROCm MFMA, and `KernelRegistry` wiring remain explicit, separately tracked follow-ups (see Known Limitations in `docs/operators/attention-varlen.md`).

**Depends on #233** (Triton varlen/LSE baseline) -- this PR is stacked on top of that branch, so its diff includes those commits until #233 merges.

## Test plan
- [x] `KERNEL_ALIGN_FORCE_SM90=1 MAX_JOBS=2 pip install -e .` builds clean on an H100 (SM90) -- `MAX_JOBS=2` needed after a full-parallelism build hit `cicc died due to signal 9` on constrained build resources.
- [x] `python -m pytest tests/test_flash_attention_sm90.py -v` -- 24/24 passing: the forward shape matrix (uneven/block-aligned/long seqlens, non-causal, decode-style, zero-length sequences, single-head, 8-entry mixed batch) now also exercises `.backward()` and checks `dQ`/`dK`/`dV`; plus dedicated single-KV-tile, multi-batch `dQ` atomic-accumulation stress, `return_lse=False`, partial `requires_grad`, non-contiguous `do`, and forward+backward input-validation error paths (bad head_dim/dtype/GQA/cu_seqlens mismatch/non-contiguous).
- [x] `python -m pytest tests/test_triton_attention_varlen.py -v` -- 9/9 passing, no regression to the baseline.
- [x] Forward output/LSE checked against both an independent fp32 reference and `triton_flash_attention_varlen`; gradients checked against the fp32 reference only (Triton's own backward can't run on bf16 tensors in this Triton version -- `tl.atomic_add` doesn't support bf16, a compiler limitation confirmed directly, not a numerical one).
- [x] Two indexing bugs (per-warp vs whole-tile loop-bound mismatch; a transpose scratch buffer indexed globally instead of locally) were caught by re-derivation before ever building, not just by test failures.

## Summary by CodeRabbit

* **New Features**
  * Added Hopper SM90 packed variable-length FlashAttention (bf16, head dim 128) with causal/non-causal support and backward gradients.
  * Added optional attention-domain LSE export (`return_lse`) for both dense and varlen paths; LSE is non-differentiable while output gradients still work.
  * Preserves automatic fallback to the Triton varlen implementation when SM90 conditions aren’t met.
* **Documentation**
  * Added operator documentation for varlen packing and LSE export behavior/contracts.
* **Tests**
  * Added/expanded correctness and gradient tests for SM90 and Triton LSE/varlen, including non-contiguous inputs and edge cases.

### DIFF
--- a/csrc/cuda/attention/flash_attention_varlen_sm90.cu
+++ b/csrc/cuda/attention/flash_attention_varlen_sm90.cu
@@ -438,6 +438,517 @@ __global__ __launch_bounds__(TB_SIZE) void flash_attention_varlen_sm90_kernel(
     }
 }
 
+// Delta[row,h] = sum_d(Out[row,h,d] * DO[row,h,d]), the per-query-row term
+// the backward recompute needs (`_bwd_preprocess_varlen` in triton_attn.py).
+// One block per (row, head); DIM=128 threads, one per element, block reduce.
+__global__ void flash_attention_bwd_preprocess_kernel(
+    const nv_bfloat16 *__restrict__ Out,
+    const nv_bfloat16 *__restrict__ DO,
+    float *__restrict__ Delta,
+    int total_q,
+    int H) {
+    const int row = blockIdx.x;
+    const int h = blockIdx.y;
+    const int tid = threadIdx.x;
+    if (row >= total_q)
+        return;
+
+    const int base = (row * H + h) * DIM;
+    float val = __bfloat162float(Out[base + tid]) * __bfloat162float(DO[base + tid]);
+
+#pragma unroll
+    for (int off = 16; off > 0; off >>= 1)
+        val += __shfl_xor_sync(0xFFFFFFFFu, val, off);
+
+    __shared__ float warp_sums[DIM / WARP_SIZE];
+    const int warp = tid / WARP_SIZE;
+    const int lane = tid % WARP_SIZE;
+    if (lane == 0)
+        warp_sums[warp] = val;
+    __syncthreads();
+    if (tid == 0) {
+        float total = 0.0f;
+#pragma unroll
+        for (int w = 0; w < DIM / WARP_SIZE; w++)
+            total += warp_sums[w];
+        Delta[row * H + h] = total;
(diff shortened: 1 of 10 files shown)


Flink-ddd:
please resolve the CI error first, Thank you.

Billy1900:
> please resolve the CI error first, Thank you.

fix it. waiting for further review.

Issue #72, opened 2026-06-03
[FEAT][observability]: Implement End-to-End Observability (NVTX Tracing + Prometheus Metrics)


Flink-ddd:
### Motivation
As the RL-Kernel and its upper-level RL-Engine scale to support large-scale distributed PPO/GRPO training, observability becomes critical. However, integrating Heavy-weight monitoring (like Prometheus) directly into the C++/CUDA operator level is an anti-pattern that introduces unacceptable overhead and breaks asynchronous CUDA streams.

We need a unified, low-overhead observability architecture that provides both **micro-level hardware tracing** (for kernel developers) and **macro-level cluster monitoring** (for SREs and researchers) via Grafana.

### Architecture Design
We propose a Two-Tiered Observability System:

1. **Micro-Level (C++ / CUDA Kernels): NVTX Tracing**
   - Inject NVIDIA Tools Extension (NVTX) markers in csrc/ops.cpp.
   - **Zero Overhead:** These markers have negligible overhead and are only activated when profiled via Nsight Systems (nsys).
   - **Goal:** Visualize exact kernel latency, GPU idle time, and stream synchronization events in flame graphs.

2. **Macro-Level (Python Engine Layer): Prometheus Metrics**
   - Integrate prometheus_client at the Python Executor/Scheduler level (rl_engine).
   - Run a lightweight background HTTP server to expose a /metrics endpoint.
   - **Goal:** Track macro metrics for Grafana dashboards, such as:
     - Output Token Throughput (Tokens/s)
     - Hardware fallback rate (e.g., when inputs don't match specific kernel shapes)
     - Rollout QPS and KV Cache fragmentation rate.

### Proposed Implementation
**Phase 1: NVTX Injection (Kernel Layer)**
- Include <nvtx3/nvToolsExt.h> in csrc/ops.cpp.
- Wrap kernel launches with nvtxRangePushA and nvtxRangePop.

**Phase 2: Prometheus Setup (Engine Layer)**
- Add prometheus-client to requirements.txt.
- Implement a MetricsRegistry singleton in Python to handle Counter, Gauge, and Histogram.
- Expose port 8000 (or configurable) for Prometheus scraping.

### Acceptance Criteria (DoD)
1. C++ operators (e.g., prefix_shared_attention) are visible as distinct colored blocks in nsys timelines.
2. The Python engine successfully exposes a /metrics endpoint.
3. A sample grafana-dashboard.json is provided in the examples/ directory.

[...]

PR #242, opened 2026-07-22
feat(observability): add NVTX tracing and Prometheus metrics

BruceLoveDecimal:
Observability: NVTX tracing + Prometheus metrics
Adds a two-layer observability stack for RL-Kernel.

NVTX tracing (micro) — rlk::<op> ranges around every C++ operator binding plus Python stage ranges (rlk::score, rlk::train_step, …), visible on an Nsight Systems timeline. Opt-in via RL_KERNEL_NVTX=1; off by default (cached branch, no cost).

Prometheus metrics (macro) — a per-process /metrics endpoint exposing token throughput, rollout QPS, stage latencies, backend/fallback state, KV-cache fragmentation, peak memory, and training loss, plus a ready-made Grafana dashboard. prometheus-client is an optional extra; without it every call is a silent no-op.

Notes

Instrumentation is exception-transparent and per-stage (never per-token). Measured overhead +0.9% on an RTX 5090.
Existing *StageResult.metrics dicts are unchanged.
Verified on GPU: CUDA extension builds, full suite green (572 passed), nsys shows the rlk::* ranges, live /metrics served.
Closes [#72](https://github.com/BruceLoveDecimal/RL-Kernel/issues/72)


nodeeeeee:
Your NVTX seems to measure the launch and return time instead of the real runtime of the kernel.

Since GPU launch and execution are asynchronous, measuring on CPU side will not reveal kernel time.

rl_engine/executors/stateless_executor.py
```diff
@@ -111,6 +114,14 @@ def __init__(
         self.reward_adapter = reward_adapter or default_reward_adapter
 
     def score(self, inputs: StatelessForwardInputs) -> StatelessForwardResult:
```
nodeeeeee, line 116: A better way is to use decorator here. See many hardcoded modifications like this

nodeeeeee (commented):
You may write a decorator as wrapper instead of using `_train_impl` and `_score_impl`.

Also, could you help enable counting op calls, in the metrics we can see how many times an op is called.

Tested on my h100 and everything seems good.

zhangj1an:
Thanks for helping me review this @nodeeeeee T-T

I will resolve conflicts and approve this PR on Tuesday (I need to check whether we need follow up PRs as the codebase has evolved a bit). Also thank you bruce for your beautiful code!!

PR #240, opened 2026-07-21
[WS1][kernels] Batch-invariant attention (standard softmax--cuda version) 

maxiaosong1124:
## Summary

  Adds a deterministic CUDA standard-softmax attention implementation for #147.

  The implementation provides batch-invariant forward and backward paths with a fixed accumulation order. It supports causal masking, key-padding masks, GQA/MQA, variable sequence lengths, chunked prefill, decode, and KV-cache handoff.

  The CUDA path matches the existing FP32 golden reference NativeAttentionOp within the shared #108 tolerance contract. Outputs and gradients are bitwise invariant when the same sequence is evaluated at different batch sizes, batch positions, or prefill chunk sizes.

  Closes #147.

  ## Implementation

  The forward path is decomposed into three deterministic CUDA kernels:

  1. QK: computes FP32 attention scores with a fixed d = 0 ... D-1 accumulation order.
  2. Masked softmax + LSE: one CTA per attention row with a fixed 256-thread shared-memory tree reduction.
  3. PV: computes the output with a fixed k = 0 ... Skv-1 FP32 accumulation order.

  The backward path implements deterministic:

  - dP
  - softmax backward
  - dQ
  - dK
  - dV

  For GQA, dK and dV accumulate query heads and query positions in a fixed order. There are no cross-CTA atomics, split-KV paths, or launch-order-dependent reductions.

   Path               Status
  ━━━━━━━━━━━━━━━━━  ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━
   CUDA forward       Deterministic QK → masked softmax/LSE → PV
  ─────────────────  ───────────────────────────────────────────────────────
   CUDA backward      Deterministic dP → softmax backward → dQ/dK/dV
  ─────────────────  ───────────────────────────────────────────────────────
   Accumulation       FP32 intermediates and statistics
  ─────────────────  ───────────────────────────────────────────────────────
   Dtypes             FP16 and BF16
  ─────────────────  ───────────────────────────────────────────────────────
   Head dimension     D=128
  ─────────────────  ───────────────────────────────────────────────────────
   Attention modes    Causal and non-causal
  ─────────────────  ───────────────────────────────────────────────────────
   Head layouts       MHA, MQA, and GQA
  ─────────────────  ───────────────────────────────────────────────────────
   Padding            key_padding_mask support, including fully masked rows
  ─────────────────  ───────────────────────────────────────────────────────
[...]

PR #246, opened 2026-07-25
feat(kernels): add Triton batch-invariant attention

### LINKED ISSUE #147: [WS1][kernels] Batch-invariant attention (standard softmax)
Part of WS1 — Full Batch-Invariant Forward Chain (epic: #<WS1 tracking issue>)

## Why

Flash-style attention parallelizes over the KV dimension and merges partial results, and the number of KV splits depends on sequence length and batch — so the online-softmax accumulation order changes with batch configuration, breaking invariance. Because attention mixes information across positions, even tiny drift here spreads to every later token.

## Scope

Provide a batch-invariant standard-softmax attention for the forward chain.

- Start from a deterministic masked-softmax + logsumexp (LSE) kernel with a fixed max / sum-exp order, then extend to the full attention output (P x V).
- Avoid split-KV / variable KV-partitioning; use a single-SM whole-sequence reduction, or a dual-kernel design with an identical, fixed accumulation order regardless of split count.
- Causal mask + padding mask support; variable-length sequence coverage.
- Keep prefill and decode on a consistent reduction path (coordinate with the KV-cache consistency issue); expose the hooks that issue needs.
- Accumulate softmax statistics in FP32.
- Validate against the #108 harness across batch=1/N, chunked-prefill on/off, padding layouts.

## Out of scope

- Non-standard attention variants (linear / sliding-window / MoE routing) — standard softmax only for this gate.
- Multi-GPU / sequence-parallel attention (WS2).
- FP8 attention.
- Full FlashAttention replacement / maximum throughput in the first deterministic baseline.

## Acceptance criteria

- The same query / sequence produces bitwise-identical (or within #108 tolerance) attention output regardless of batch size, batch position, and chunked-prefill on/off.
- No split-KV path whose split count varies with batch / seq config; the accumulation order is pinned and documented.
- Softmax + LSE outputs are invariant across the sweep; tests cover both fixed-length and variable-length sequences.
- Decode-stage output matches the prefill path for the same effective context (handoff with the KV-cache issue).
- The attention backward path passes the shared gradient-invariance check from the WS1 backward-consistency issue.

## Notes

- Tightly coupled with the KV-cache path-consistency issue — agree on the shared reduction contract early.
- Existing attention kernels (#65 prefix-shared fused attention, #41 Triton FlashAttention fallback) are reference / audit starting points, not the invariant path itself.
- Single-SM whole-sequence is simplest to prove correct but may cost throughput on long sequences; note the tradeoff if choosing the dual-kernel route.

## Planned PRs

- [ ] Deterministic masked-softmax + LSE kernel (fixed max / sum-exp order)
- [ ] Causal + padding mask tests; variable-length coverage
- [ ] Deterministic P x V aggregation (single-SM or fixed-order dual-kernel)
- [ ] Prefill/decode handoff hooks (coordinate with the KV-cache issue)
- [ ] Wire through the #108 harness sweep; benchmark + tradeoff note

EthanZero2Hero:
Closes #147.

This PR adds a Triton standard-softmax attention baseline for WS1 batch-invariant forward validation. The kernel uses a fixed key-order reduction for max, sum-exp/LSE, and P @ V, supports causal and key-padding masks, and is wired into the #108 operator harness as the attention/triton candidate.

## Summary by CodeRabbit

* **New Features**
  * Added a Triton-accelerated attention operator supporting causal and non-causal modes, configurable scaling, key padding masks, and grouped-query attention.
  * Added an API variant that returns both attention output and log-sum-exp (LSE) statistics.

* **Tests**
  * Added CUDA-gated test coverage for forward parity vs the native reference, LSE correctness, padding-layout and batch-position invariance, decode vs. prefill equivalence, kv-cache handoff behavior, and backward/gradient correctness.

* **Chores**
  * Updated attention-specific accuracy tolerances used for kernel verification, and extended the tolerance contract to include the new attention operator.

### DIFF
--- a/rl_engine/kernels/gtest/operator_specs.py
+++ b/rl_engine/kernels/gtest/operator_specs.py
@@ -32,6 +32,20 @@ def _load_object(path: str) -> Any:
 
 
 OP_SPECS = {
+    "attention": OperatorSpec(
+        name="attention",
+        op_class="attention",
+        gold_path="rl_engine.kernels.ops.pytorch.attention.standard_attn.NativeAttentionOp",
+        gold_method="forward_fp32",
+        candidate_paths={
+            "pytorch": "rl_engine.kernels.ops.pytorch.attention.standard_attn.NativeAttentionOp",
+            "triton": (
+                "rl_engine.kernels.ops.triton.attention.standard_attn."
+                "TritonBatchInvariantAttentionOp"
+            ),
+        },
+        grad_input_names=("q", "k", "v"),
+    ),
     "logp": OperatorSpec(
         name="logp",
         op_class="logprob",
--- a/rl_engine/kernels/ops/triton/attention/standard_attn.py
+++ b/rl_engine/kernels/ops/triton/attention/standard_attn.py
@@ -1,0 +1,359 @@
+# SPDX-License-Identifier: Apache-2.0
+# Copyright (c) 2026 RL-Kernel Contributors
+
+from __future__ import annotations
+
+import math
+from typing import Optional
+
+import torch
+import triton
+import triton.language as tl
+
+from rl_engine.kernels.ops.pytorch.attention.standard_attn import NativeAttentionOp
+
+_BLOCK_N = 64
+
+
+def _next_power_of_2(value: int) -> int:
+    return 1 << (value - 1).bit_length()
+
+
+@triton.jit
+def _standard_attn_fwd_kernel(
+    q_ptr,
+    k_ptr,
+    v_ptr,
+    mask_ptr,
+    out_ptr,
+    lse_ptr,
+    B: tl.constexpr,
+    H_Q: tl.constexpr,
+    H_KV: tl.constexpr,
+    S_Q: tl.constexpr,
+    S_KV: tl.constexpr,
+    D: tl.constexpr,
+    stride_qb: tl.constexpr,
+    stride_qh: tl.constexpr,
+    stride_qs: tl.constexpr,
+    stride_qd: tl.constexpr,
+    stride_kb: tl.constexpr,
+    stride_kh: tl.constexpr,
+    stride_ks: tl.constexpr,
+    stride_kd: tl.constexpr,
+    stride_vb: tl.constexpr,
+    stride_vh: tl.constexpr,
+    stride_vs: tl.constexpr,
+    stride_vd: tl.constexpr,
+    stride_ob: tl.constexpr,
+    stride_oh: tl.constexpr,
+    stride_os: tl.constexpr,
+    stride_od: tl.constexpr,
+    sm_scale: tl.constexpr,
+    BLOCK_N: tl.constexpr,
+    BLOCK_D: tl.constexpr,
+    CAUSAL: tl.constexpr,
+    HAS_KEY_PADDING_MASK: tl.constexpr,
+):
+    row = tl.program_id(0)
+    q_head = tl.program_id(1)
+    batch = tl.program_id(2)
+    kv_head = q_head // (H_Q // H_KV)
+
+    offs_d = tl.arange(0, BLOCK_D)
+    d_mask = offs_d < D
+    q = tl.load(
+        q_ptr + batch * stride_qb + q_head * stride_qh + row * stride_qs + offs_d * stride_qd,
+        mask=d_mask,
+        other=0.0,
+    ).to(tl.float32)
+
+    max_score = -float("inf")
+    for start_n in range(0, S_KV, BLOCK_N):
+        cols = start_n + tl.arange(0, BLOCK_N)
+        col_mask = cols < S_KV
+        k = tl.load(
+            k_ptr
+            + batch * stride_kb
+            + kv_head * stride_kh
+            + cols[:, None] * stride_ks
+            + offs_d[None, :] * stride_kd,
+            mask=col_mask[:, None] & d_mask[None, :],
+            other=0.0,
+        ).to(tl.float32)
+        scores = tl.sum(k * q[None, :], axis=1) * sm_scale
+        scores = tl.where(col_mask, scores, -float("inf"))
+
+        if CAUSAL:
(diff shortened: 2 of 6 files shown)


rl_engine/kernels/ops/triton/attention/standard_attn.py
```diff
@@ -0,0 +1,358 @@
...
+        lse = torch.empty((batch, q_heads, q_len), device=q.device, dtype=torch.float32)
+        block_d = _next_power_of_2(head_dim)
+        grid = (q_len, q_heads, batch)
+        dummy_mask = (
+            key_padding_mask
+            if key_padding_mask is not None
+            else q.new_empty((1,), dtype=torch.bool)
+        )
+
+        _standard_attn_fwd_kernel[grid](
```
Flink-ddd, line 189: Add ctx.mark_non_differentiable(lse) before the return statement when return_lse is True, so PyTorch's autograd engine doesn't expect a gradient for the LSE tensor (since your custom backward explicitly ignores grad_lse).

rl_engine/kernels/gtest/operator_specs.py
```diff
@@ -32,6 +32,20 @@ def _load_object(path: str) -> Any:
 
 
 OP_SPECS = {
+    "attention": OperatorSpec(
+        name="attention",
+        op_class="reduction",
```
Flink-ddd: Change op_class="reduction" to op_class="attention" to ensure the test harness maps this Triton candidate to the correct attention tolerance contract established in PR #240.

Flink-ddd (changes requested):
Left a few comments, otherwise LGTM!

rl_engine/kernels/ops/triton/attention/standard_attn.py
EthanZero2Hero: Thanks, fixed.

I added `ctx.mark_non_differentiable(lse)` before returning `(out, lse)` when `return_lse=True`, so autograd will not expect or propagate a gradient for the LSE tensor. I also added a regression test that verifies `lse.requires_grad is False` while gradients for q/k/v still work through the attention output.

rl_engine/kernels/gtest/operator_specs.py
EthanZero2Hero: Thanks, addressed.

Changed the attention operator spec from `op_class="reduction"` to `op_class="attention"` so the #108 harness uses the attention-specific tolerance contract.

I also added the `attention` tolerance entry locally, matching the contract from PR #240, so this PR can run the harness cleanly on the current base.

tests/test_triton_batch_invariant_attention.py
```diff
@@ -0,0 +1,235 @@
...
+def test_triton_attention_decode_matches_prefill_suffix_context():
+    dtype = torch.bfloat16
+    q_full, k_full, v_full = _qkv(1, 12, 12, dtype=dtype, seed=6)
+    op = TritonBatchInvariantAttentionOp()
+
+    prefill = op(q_full, k_full, v_full, causal=True)
+    decode = op(q_full[:, :, -1:], k_full, v_full, causal=True)
+
+    assert torch.equal(prefill[:, :, -1:], decode)
+
```
maxiaosong1124, line 259: Please add coverage equivalent to the CUDA tests for:
  - full prefill vs sequence-dimension chunked prefill with several chunk sizes;
  - chunked-prefill LSE invariance;
  - chunked prefill with key_padding_mask;
  - KV-cache handoff with S_new=1 and S_new>1;
  - representative MHA/GQA configurations.

  The comparisons for equivalent contexts should use torch.equal where the
  batch-invariance contract requires bitwise identity.

tests/test_triton_batch_invariant_attention.py
```diff
@@ -0,0 +1,235 @@
...
+    out.backward(dy)
+    return out.detach(), q_req.grad.detach(), k_req.grad.detach(), v_req.grad.detach()
+
+
+@requires_cuda
+@pytest.mark.parametrize("causal", [False, True])
+@pytest.mark.parametrize(
+    "dtype, atol, rtol", [(torch.float16, 3e-3, 3e-3), (torch.bfloat16, 5e-2, 2e-2)]
+)
+def test_triton_attention_matches_native_forward(causal, dtype, atol, rtol):
```
maxiaosong1124, line 68: Please add an all-False key_padding_mask row and verify:
  - output is finite and exactly zero;
  - LSE is -inf;
  - behavior matches the NativeAttentionOp convention;
  - backward produces finite gradients and no contribution from the fully masked
    row.

maxiaosong1124 (commented):
Left some comments,please resolve these problems, thanks!

tests/test_triton_batch_invariant_attention.py
EthanZero2Hero: Thanks, added Triton coverage equivalent to the CUDA batch-invariance tests.

This update adds:
- full prefill vs sequence-dimension chunked prefill across several chunk sizes;
- chunked-prefill LSE invariance;
- chunked prefill with `key_padding_mask`;
- KV-cache handoff coverage for both `S_new=1` and `S_new>1`;
- representative MHA/GQA configurations.

For equivalent contexts, the tests now use `torch.equal` for both output and LSE to enforce the bitwise batch-invariance contract.

tests/test_triton_batch_invariant_attention.py
EthanZero2Hero: Thanks, added an all-False `key_padding_mask` row test.

The test verifies:
- the fully masked row output is finite and exactly zero;
- LSE is `-inf`;
- the behavior matches `NativeAttentionOp`;
- backward produces finite gradients;
- the fully masked row contributes exactly zero gradients for q/k/v.

Flink-ddd (approved):

maxiaosong1124 (approved):
LGTM now! thanks for your update!

PR #250, opened 2026-07-27
feat(observability): NVTX kernel tracing + Prometheus metrics endpoint

Billy1900:
## Summary
- Adds NVTX ranges around every operator bound in `csrc/ops.cpp` (via a small `csrc/utils/nvtx_utils.h` RAII helper) so each RL-Kernel op shows up as a distinct, named block in an `nsys` timeline instead of an undifferentiated stream of raw kernel launches.
- Adds an opt-in Prometheus `/metrics` HTTP endpoint (`rl_engine/observability/metrics.py`) tracking kernel-op throughput/latency, backend-fallback rate, and paged-KV-cache fragmentation, wired into `KernelRegistry.get_op`, `paged_kv_baseline.collect_paged_kv_metrics`, and `RolloutExecutor`.
- Ships a sample Grafana dashboard (`examples/grafana/rl_kernel_dashboard.json`) and two new docs pages (`nsys-profiling.md`, `metrics-and-dashboards.md`).

Addresses #72.

## Design notes
- `RL_KERNEL_ENABLE_OP_METRICS` (op-latency wrapping) defaults **off**: several existing tests assert `isinstance(kernel_registry.get_op(...), ConcreteOpClass)`, and wrapping by default would break them. Backend-fallback and KV-cache-fragmentation recording require no opt-in beyond having `prometheus_client` installed (`pip install -e .[observability]`), since they never change any function's return type.
- NVTX links against `libnvToolsExt` (`-lnvToolsExt` added to `setup.py`'s CUDA build) rather than relying on `nvtx3`'s dlopen-based injection layer, since the classic `<nvToolsExt.h>` API used here resolves its symbols at link time.
- `start_metrics_server` catches port-bind failures (`OSError`) and degrades to returning `None` rather than crashing a worker — observability must never take down real training/rollout.

## Test plan
- [x] `pytest tests/test_observability_metrics.py -v` — 15/15 passing (CPU-only; no-op-without-dependency, `InstrumentedOp` call/latency recording including on exceptions, registry fallback recording including full-exhaustion, KV-cache fragmentation, `/metrics` endpoint smoke, port-conflict handling, malformed-env-var handling)
- [x] Full suite (`pytest tests/ -q`): 620 passed, 1 pre-existing failure unrelated to this change (confirmed identical on unmodified `main`)
- [x] Rebuilt the CUDA extension on real H100 hardware and ran `nsys profile` + `nsys stats --report nvtx_sum`: confirmed labeled `rl_kernel::fused_logp` / `rl_kernel::deterministic_logp` ranges, including verifying the NVTX RAII range still closes correctly (no orphaned ranges) when the wrapped op raises
- [ ] Manual: point a local Prometheus/Grafana at a worker's `/metrics` endpoint and import the sample dashboard

## Summary by CodeRabbit

* **New Features**
  * Added optional Prometheus metrics for kernel call counts, latency, backend fallbacks, and KV-cache fragmentation.
  * Added an HTTP `/metrics` endpoint with opt-in metrics server startup.
  * Added a Grafana “RL-Kernel Observability” dashboard template.
  * Added NVTX range annotations to improve CUDA operator visibility in Nsight Systems.
* **Documentation**
  * Added “Metrics & Dashboards” and “NVTX & Nsight Profiling” guides.
* **Tests**
  * Added CPU-only tests covering metrics gating, instrumentation, fallback tracking, KV fragmentation, and metrics server behavior.

### DIFF
--- a/csrc/ops.cpp
+++ b/csrc/ops.cpp
@@ -4,6 +4,8 @@
 #include <torch/extension.h>
 #include <cuda_bf16.h>
 
+#include "utils/nvtx_utils.h"
+
 // Fused LogP Declarations
 torch::Tensor fused_logp_forward(torch::Tensor logits, torch::Tensor token_ids);
 
@@ -93,6 +95,7 @@ at::Tensor prefix_shared_attention(
   const at::Tensor& K,
   const at::Tensor& V)
 {
+  RL_KERNEL_NVTX_RANGE("rl_kernel::prefix_shared_attention");
   TORCH_CHECK(Q.dim() == 4, "Q must be [bs, G, len_q, DIM]");
   TORCH_CHECK(K.dim() == 3, "K must be [bs, len_kv, DIM]");
   TORCH_CHECK(V.dim() == 3, "V must be [bs, len_kv, DIM]");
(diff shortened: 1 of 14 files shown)


csrc/ops.cpp
nodeeeeee: The instrumentation is still enabled when we disable profiling. This might cause some performance loss.

PR #264, opened 2026-08-05
docs(ws1): batch-invariance contract RFC (#101)

### LINKED ISSUE #101: [RFC] Batch-Invariant RL Kernel Suite for Train-Inference Consistency
This RFC defines the batch-invariance contract for RL-Kernel's train-inference consistency suite.

The target identity is:

```text
same model + same token sequence + same policy state
=> aligned logprobs / hidden states regardless of surrounding batch shape
```

This is the P0.3 roadmap item from #83. It is broader than the CUDA deterministic logprob kernel in #96: #96 is one concrete operator path, while this RFC should define the cross-kernel and cross-engine contract that rollout and training backends must satisfy.

## Problem

RL rollout and training rarely execute the same sequence in the same shape:

- rollout engines use dynamic batching, chunked prefill, paged KV, and prefix cache;
- reference/reward scoring may use different packing and padding layouts;
- training engines use micro-batches, gradient accumulation, FSDP/TP partitioning, and different row order;
- kernel dispatch may choose different CUDA/ROCm/Triton implementations by shape or hardware.

Small logprob drift can be amplified by GRPO/PPO ratio and KL terms. The suite should make drift visible and locate whether it comes from logprob reduction, attention/prefix reuse, packing layout, TP reduction, dtype rounding, or engine integration.

## Scope

Define the expected invariance contract for the same logical sequence across:

- batch size changes: B=1 / small batch / large batch;
- batch position changes and unrelated neighboring samples;
- chunked prefill on/off;
- prefix cache on/off;
- padding layout and varlen packing changes;
- dense vs indexed selected-logprob paths;
- CUDA / ROCm / Triton backend parity, with backend-specific tolerance policy where bitwise identity is not realistic;
- rollout engines such as vLLM/sglang vs training engines such as Megatron/DeepSpeed/FSDP.

## Questions To Resolve

- Which outputs require bitwise identity, and which require explicit tolerance?
- What is the canonical fixture format for prompt, completion, token ids, position ids, masks, cache metadata, and policy state?
- Which kernels are included in the first suite: selected logprob, attention with exported LSE, prefix-shared attention, sampling, loss reductions?
- How should failures report the first divergent layer/operator?
- How should TP>1 reductions compare against FSDP(TP=1)?
- What command should contributors run locally before opening PRs?
- Which checks belong in CI, and which require GPU/nightly/benchmark lanes?

## Proposed Deliverables

- RFC document describing the invariance contract and tolerance policy.
- A reusable fixture generator for logically identical sequences under different batch/cache/layout configurations.
- A batch-invariance test matrix for selected logprob and attention-adjacent paths.
- A cross-engine benchmark plan comparing rollout logprobs with training logprobs.
- A drift report format that includes max error, sequence id, token position, layer/operator, backend, dtype, and launch configuration.
- Links to implementation issues for concrete kernels and tooling.

## Acceptance Criteria

- The RFC clearly separates bitwise-invariant requirements from tolerance-based parity requirements.
- The RFC names the minimal P0 suite and the deferred P1/P2 extensions.
- The RFC references #96 as the CUDA deterministic selected-logprob implementation path.
- The RFC defines at least one local command shape for contributors and one CI/nightly command shape for maintainers.
- Follow-up issues can be opened directly from the RFC without re-litigating the contract.

Billy1900:
## Summary

Drafts the RFC requested by #101: the batch-invariance contract for RL-Kernel's train-inference consistency suite.

Filed as **WS1** rather than a new workstream — WS1 in #83 already is "full batch-invariant forward chain." This is the missing contract document over WS1's existing implementation issues, plus the localisation half of WS4.

## The core proposal

Three comparison relations, so the threshold for any check follows from its kind rather than from case-by-case negotiation:

| Relation | Varies | Threshold |
| --- | --- | --- |
| **I** (invariance) | batch shape only | **bitwise (0, 0)** |
| **A** (accuracy) | impl vs fp32 gold | #108 accuracy table |
| **P** (parity) | backend vs backend | #108 accuracy table |

The argument for bitwise on I: the kernel, dtype, data and device are all held fixed and only the surrounding shape moves, so there is no legitimate source of floating-point difference. A tolerance there would hide exactly the reduction-schedule defect the suite exists to find. WS2 correctly refuses this stance for cross-engine comparison; WS1 can demand it.

Consequence worth noting: **an I-relation needs no reference implementation** — the kernel is its own oracle. That is what makes the P0 suite single-device, seconds-long, and viable as a PR gate where WS2 structurally is not.

## Two findings that shaped the document

1. `tolerance_contract.json` already declares `"batch_invariance": {"atol": 0.0, "rtol": 0.0}`, and **nothing reads it** — `op_checks.py::_resolve_tolerance` only consults `contract["accuracy"]`. Five test files independently carry a copy-pasted `Axis A -- batch invariance, bitwise` comment. The policy is settled in practice; this RFC ratifies and wires it rather than proposing it.
2. `run_operator_suite` implements the A-relation shape (two functions, one input). The I-relation is one function, two input layouts, exact comparison — different enough to warrant a sibling `run_invariance_suite` rather than a flag.

## Also included

- Six transform classes (T1 population, T2 position, T3 neighbour content, T4 padding, T5 execution schedule, T6 selection/packing), named as `source_class` values compatible with the WS2 `KnobDefinition` model so a WS1 axis lifts into a WS2 ablation without redefinition.
- `InvarianceDeclaration | None` on `OperatorSpec` with a **required `exempt_reason`** when null (`NativeGemmOp` -> "non-deterministic reference baseline"), making the matrix self-describing.
- Drift report schema and chain-index fault localisation.
- P0 suite matrix, with T7 prefix-cache, prefix-shared attention, ROCm and sampling deferred to P1 with stated reasons.
- A measured coverage table of what exists today. Coverage is real but scattered across ~30 hand-written tests in a dozen files; `grpo_loss` and `ratio_kl` have none.

## Positioning

WS1 is a precondition for WS2, not a competitor. The WS2 ablation ladder starts at `A0: fully aligned reference`; if the underlying kernels are not batch-invariant, A0 is nondeterministic under batching and everything above it is noise.

## Open question for reviewers

The RFC asserts **bitwise as the P0 requirement** rather than leaving it open. That is the decision most worth challenging: if invariance is tolerance-based instead, the suite collapses into a WS2 knob and the separate PR-blocking CI lane loses its justification. The evidence says bitwise is already what the codebase does.

No implementation in this PR — the roadmap table lists 12 follow-up work items, deliberately left unfiled so maintainers can decide how to slice them.

Closes #101 as an RFC deliverable.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

## Summary by CodeRabbit

* **Documentation**
  * Added an RFC defining batch-shape invariance requirements for reinforcement learning kernels.
  * Documented supported transformations, accuracy and backend-parity expectations, active-token handling, deterministic serialization, and drift reporting.
  * Added guidance for fixtures, test coverage levels, command-line and CI integration, failure classification, replay support, and completion criteria.

### DIFF
--- a/docs/design/ws1_batch_invariance_contract.md
+++ b/docs/design/ws1_batch_invariance_contract.md
@@ -1,0 +1,541 @@
+# WS1 Batch-Invariance Contract
+
+Status: RFC
+
+Tracking issues:
+
+- [#101: Batch-invariant RL kernel suite for train-inference consistency](https://github.com/RL-Align/RL-Kernel/issues/101)
+- [#83: RL-Kernel roadmap Q3-Q4](https://github.com/RL-Align/RL-Kernel/issues/83)
+- [#108: WS1 numerical contract](https://github.com/RL-Align/RL-Kernel/issues/108)
+- [#96: Batch-invariant deterministic logprob CUDA kernel](https://github.com/RL-Align/RL-Kernel/issues/96)
+- [#111: WS2 cross-config alignment](https://github.com/RL-Align/RL-Kernel/issues/111)
+
+## Motivation
+
+RL post-training requires that the same token sequence score the same way regardless of what
+else happens to be in the batch. Rollout engines use dynamic batching, chunked prefill, paged
+KV, and prefix caching. Reference and reward scoring use different packing and padding. Training
+uses micro-batches, gradient accumulation, and different row order. None of these are supposed to
+change the logprob of a given token, and all of them can.
+
+The target identity is:
+
+```text
+same model + same token sequence + same policy state
+    => identical active-token outputs, regardless of surrounding batch shape
+```
+
+This is a different failure axis from cross-engine drift. WS2 ([#111]) measures drift between two
+*execution paths* — rollout vs training — and its primary metric is
+`dlogp = train_recomputed_logp - rollout_old_logp`. WS1 measures drift within a *single* path when
+only the batch shape around a sequence changes.
+
+The relationship is a dependency, not an overlap. The WS2 ablation ladder begins at `A0: fully
+aligned reference` and states that if A0 fails, the failure is not yet a cross-config problem. A0
+is only meaningful if the underlying kernels are batch-invariant. If they are not, A0 is
+nondeterministic under batching and every WS2 measurement stacked above it is noise. **This RFC
+defines the contract that makes A0 trustworthy.**
+
+Batch invariance is also the axis where the strictest possible guarantee is actually achievable.
+When only the batch shape changes — same kernel, same dtype, same reduction tree, same device —
+there is no legitimate source of floating-point difference. Any difference is a bug in the
+reduction schedule, not an inherent cost of the hardware. WS1 therefore demands bitwise identity
+where WS2 correctly refuses to.
+
+## Relationship To Adjacent Contracts
+
+| Contract | Compares | Axis | Requirement |
+| --- | --- | --- | --- |
+| WS1 accuracy ([#108]) | candidate kernel vs fp32 gold | implementation quality | per-dtype tolerance |
+| **WS1 invariance (this RFC)** | **kernel vs itself under a shape transform** | **batch shape** | **bitwise** |
+| WS2 cross-config ([#111]) | rollout path vs training path | engine / parallelism | #108 `logprob` tolerance |
+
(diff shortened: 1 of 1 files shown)


Flink-ddd:
cc @maxiaosong1124 @frank-2077 PTAL, Do we need this PR?

PR #310, opened 2026-08-15
feat(distributed): add deterministic TP8 all-reduce

CyberSecurityErial:
## Summary by CodeRabbit

* **New Features**
  * Added a deterministic CUDA collective for single-host workloads with world sizes of 1, 2, 4, and 8.
  * Supports deterministic all-reduce operations for float32, float16, and bfloat16 tensors.
  * Added optional output buffers and context-manager resource handling.
  * Added validation for devices, tensor layouts, sizes, data types, and distributed configuration.
* **Tests**
  * Added coverage for repeated reductions, cross-rank consistency, supported data types, and output-buffer behavior across supported world sizes.


rl_engine/distributed/collectives.py
```diff
@@ -0,0 +1,234 @@
...
+                "the RL-Kernel CUDA extension lacks deterministic collectives: "
+                + ", ".join(missing)
+            )
+
+        self.device = normalized_device
+        self.max_size_bytes = int(max_size_bytes)
+        self._extension = _C
+        self._lock = threading.Lock()
+        self._handle = 0
+        self._staging = torch.empty(
```
Flink-ddd, line 99: Could we use a dedicated IPC-safe cudaMalloc allocation for this staging buffer instead of a PyTorch caching-allocator tensor? cudaIpcGetMemHandle is obtained from the allocation base, which may represent a larger allocator segment and can expose or mishandle unrelated memory.

Flink-ddd (commented):
I left a comment, other LGTM. Thank you for excellent work!

rl_engine/distributed/collectives.py
```diff
@@ -0,0 +1,234 @@
...
+
+    def close(self) -> None:
+        """Release imported CUDA IPC mappings after the last collective call."""
+
+        handle = getattr(self, "_handle", 0)
+        if not handle:
+            return
+        torch.cuda.synchronize(self.device)
+        self._handle = 0
+        self._extension.deterministic_collective_destroy(handle)
```
zhangj1an, line 169: (Optional) The existing close() only ensures that the local GPU has finished its work. It may be safer to close the IPC resources only after ensuring that no peer GPU is still reading this GPU's staging buffer.

rl_engine/distributed/collectives.py
zhangj1an: +1, This Zhihu CUDA IPC tutorial by Kaiyuan also uses `cudaMalloc` and then `cudaIpcGetMemHandle`,
https://zhuanlan.zhihu.com/p/2019510762004050171, so it looks safe to follow this pattern

zhangj1an (approved):
LGTM, is good to merge. 

This PR uses CUDA IPC,
Image (image taken from [here](https://zhuanlan.zhihu.com/p/2019510762004050171))

the all reduce order is
```
TP=1 : r0

TP=2 : sum01
       = r0 + r1

TP=4 : sum03
       = (r0+r1) + (r2+r3)

TP=8 : sum07
       = ((r0+r1)+(r2+r3))
       + ((r4+r5)+(r6+r7))
```


the collective lifecycle is `stage -> sync -> fixed-tree all-reduce -> sync -> ...`.

For simplicity, barrier is good enough, no need to use `cudaStreamWaitEvent(peer_ready_event)`.

Flink-ddd (approved):
LGTM

PR #312, opened 2026-08-15
feat(distributed): add deterministic TP8 all-gather

CyberSecurityErial:


zhangj1an:
Can we replace this PR with NCCL AllGather? ( `dist.all_gather_into_tensor(out, input, group=self.group)`) Because AllGather does not involve floating-point reduction

Flink-ddd (approved):
LGTM, Thank you for excellent work!

inaniloquentee (approved):

Issue #335, opened 2026-08-24
[Vime][TP=2 CP=2] Strict R/R linear_logp remains nonzero (0.023694925) after shared backend


inaniloquentee:
## Summary

We ran Vime R/R (rollout/reference versus training/reference) on 8x H100 with Qwen3-8B, CP=2, TP=2. Both rollout and training report the same deterministic RL-Kernel operator providers, including the separate strict `linear_logp` backend:

```text
attention:  rlkernel.attention.deterministic.v1
ffn:        rlkernel.ffn.qwen3.deterministic.v1
linear_logp: rlkernel.linear_logp.bitwise.v1
```

The end-to-end job completed rollout and training and wrote an Nsight trace, but failed Vime's zero-threshold assertion:

```text
train/train_rollout_logprob_abs_diff = 0.023694925010204315
threshold = 0
```

This is **not** a strict-bitwise pass. The metric above is Vime's `sum_of_sample_mean(abs(...))`; it is not the maximum per-element absolute difference.

The immediate goal is to locate and eliminate the remaining difference source so R/R satisfies elementwise `mismatch_count == 0`, `max_abs_diff == 0`, and the Vime aggregate is exactly zero.

## Environment and Reproduction

| Item | Value |
| --- | --- |
| Host | `xe9680-cn-2` |
| GPU | 8x H100 |
| RL-Kernel checkout | `/home/ellm/ljj/RL-Kernel-pr230-final` |
| RL-Kernel branch/base | `codex/pr230-h100-final` / `2717867` |
| Vime checkout | `/home/ellm/ljj/vime` |
| Ray job | `raysubmit_EaaERakEkuwF1G5S` |
| Run directory | `/home/ellm/ljj/vime/outputs/rlkernel/rr-20260823-235754-strict` |
| Launcher | `/home/ellm/ljj/vime/scripts/run-qwen3-8B-rlkernel-tp2-cp2-trace.sh` |
| Nsight trace | `/home/ellm/ljj/vime/outputs/rlkernel/rr-20260823-235754-strict/trace/qwen3-8b-tp2-cp2.nsys-rep` |
| Ray driver log | `/tmp/ray/session_latest/logs/job-driver-raysubmit_EaaERakEkuwF1G5S.log` |

[...]

PR #336, opened 2026-08-24
feat(linear-logp): add strict bitwise SM90 contract

inaniloquentee:
## Summary

- freeze the SM90 fused linear_logp reduction contract so vocab splitting and reduction order no longer depend on batch shape, occupancy, or SM count
- apply temperature at the same FP32 epilogue point, optionally return sampler logits, and mask padded vocabulary lanes out of the LSE
- expose separate strict single-rank and tensor-parallel entry points while preserving the existing fast path
- make TP LSE and selected-logit merges explicitly rank ordered and keep strict autograd mathematically correct

Refs #335

## Validation

- built the CUDA extension from source for CUDA 12.4 / SM90 with `KERNEL_ALIGN_FORCE_SM90=1`
- `python -m pytest tests/test_linear_logp.py -q` (`41 passed`)
- strict tests compare raw FP32 bit patterns across full-batch and single-row execution, `return_logits` modes, temperature, and padded-vocab masking
- `flake8` on changed Python files and `git diff --check`


Flink-ddd (approved):
LGTM

maxiaosong1124 (approved):
LGTM!

PR #367, opened 2026-08-30
Enable exact-batch CUDA Graph for strict rollout

inaniloquentee:
## Summary

This PR updates the strict R/R path for the exact-bitwise contract and is ready to merge into main. It combines:

1. A deterministic TP reduction path for strict vLLM o_proj, while preserving the fixed arithmetic/reduction order.
2. Automatic exact-batch vLLM FULL_DECODE_ONLY CUDA Graph capture for strict rollout.

The change is limited to RL-Kernel integration and kernels. No Vime framework-layer source code is modified.

## Why

The trace shows that strict R/R is slower than P/P for two separate reasons:

- Strict deterministic TP collectives must preserve a fixed reduction order and therefore cannot use the faster production/custom all-reduce path.
- In eager decode, every generated token repeatedly pays host launches, stage waits, sequence publication, deterministic collective launches, and synchronization across the model layers.

CUDA Graph removes the repeated eager launch and synchronization gaps while replaying the same deterministic kernels. It does not change the reduction tree, collective protocol, stage order, or arithmetic, so it preserves the strict bitwise contract.

## Implementation

- Route strict vLLM RowParallelLinear o_proj reductions through RL-Kernel deterministic all-reduce.
- Keep the fixed-tree deterministic collective protocol; custom all-reduce and unsafe fusion remain disabled.
- Detect strict mode from --linear-logp-provider-mode strict.
- Derive the maximum graph capture size from --rollout-batch-size multiplied by --n-samples-per-prompt.
- Capture every exact batch size from 1 through the maximum, avoiding sparse-graph padding that can violate strict custom-kernel assumptions.
- Automatically pass FULL_DECODE_ONLY and --vllm-optimization-level 0.
- Respect explicit --vllm-enforce-eager, --vllm-optimization-level, and --vllm-compilation-config arguments.
- Support RL_KERNEL_VLLM_CUDAGRAPH_MAX_CAPTURE_SIZE as an optional override.
- Keep Inductor/fusion disabled in the strict path.

## Validation

Environment: Qwen3-8B, TP=2, CP=2, 4 actor GPUs + 4 rollout GPUs, global batch 8, max tokens/GPU 2048, one rollout, identical seeds, deterministic inference, custom all-reduce disabled, and disk weight transport.

| Case | Rollout | Actor train | Step |
| --- | ---: | ---: | ---: |
| P/P, Graph | 5.157s | 5.414s | 51.619s |
| R/R, Graph, previous run | 6.032s | 9.695s | 57.975s |
| R/R, Graph, latest run | 5.894s | 9.694s | 57.943s |
| P/P, eager | 19.713s | 5.359s | 64.396s |
| R/R, eager, current fixed-tree | 32.418s | 9.697s | 80.671s |

The latest R/R Graph run compared with the previous R/R Graph run changes rollout by -0.138s and the full step by -0.032s. This is not an additional seconds-level gain over an already-enabled Graph configuration.

The important seconds-level result is the comparison with the current fixed-tree eager R/R run: the latest Graph run reduces rollout by about 26.524s and the full step by about 22.728s. Without Graph, current strict R/R is 12.705s slower in rollout, 4.338s slower in actor train, and 16.275s slower per step than P/P eager. With both cases using Graph, R/R remains about 0.737s slower in rollout and 6.324s slower per step than P/P.

Without Graph, strict R/R is expected to remain substantially slower than P/P under the current bitwise contract. This is not a proven theoretical lower bound: further optimization is possible only if it preserves the same reduction tree, ordering, and synchronization semantics. The current safe optimization targets the dominant eager launch overhead; arbitrary collective fusion or custom all-reduce would violate the contract or regress performance.

## Correctness

- Latest one-round R/R Graph job completed successfully.
- Strict rollout logprob mismatch count: 0.
- Strict rollout logprob max absolute difference: 0.
- Debug train data comparison across all four ranks: total_mismatch_count=0.
- Both rollout engines captured exact decode graphs for batch sizes [1,2,3,4,5,6,7,8].
- The deterministic collective source was not changed by the rejected fused fast path; that experiment regressed Graph rollout by about 1.9s and was removed.

## Summary by CodeRabbit

- **Performance**
  - Improved deterministic distributed reductions for small payloads with a faster execution path.
  - Improved graph-replayed fused operations by using a safer staged reduction path.
  - Added deterministic tensor-parallel output reductions for supported vLLM models.

- **Startup and Compatibility**
  - Added automatic CUDA Graph configuration for strict decoding workloads when settings are not explicitly provided.
  - Added optional precompilation of attention training kernels to reduce runtime compilation delays.
  - Added validation and clear fallback warnings for unsupported or incomplete GPU configurations.

### DIFF
--- a/csrc/cuda/distributed/deterministic_collective.cu
+++ b/csrc/cuda/distributed/deterministic_collective.cu
@@ -949,7 +949,10 @@ class DeterministicCollectiveState {
     has_staged_input_ = true;
   }
 
-  void all_reduce(torch::Tensor& output, cudaStream_t stream) {
+  void all_reduce(
+      torch::Tensor& output,
+      cudaStream_t stream,
+      bool allow_owner_path = true) {
     check_tensor(output, "output");
     TORCH_CHECK(has_staged_input_, "stage() must be called before all_reduce()");
     TORCH_CHECK(
@@ -960,7 +963,10 @@ class DeterministicCollectiveState {
         "all-reduce output size must match the staged input size");
 
     const int64_t element_count = output.numel();
-    if (staged_owner_path_) {
+    // Graph-replayed calls must avoid the owner-push branch because it writes
+    // to remote IPC frames. Callers that use the fused ABI pass false here;
+    // direct staged collectives retain the eager owner optimization.
+    if (allow_owner_path && staged_owner_path_) {
       if (rank_ == 0) {
         // Logical rank 0 is the topology-favorable reader in both traced TP
         // engines. It evaluates the original fixed tree exactly once, then
@@ -1090,10 +1096,11 @@ class DeterministicCollectiveState {
         output.numel() == input.numel(),
         "all-reduce output size must match the input size");
 
-    // Route both ABI entry points through the same staged protocol so the
-    // topology-aware owner reduction is always applied.
+    // Use the graph-safe staged protocol for every message size. The fused
+    // two-slot protocol is intentionally kept available in the extension for
+    // experiments, but is not safe to replay across vLLM's many graph shapes.
     stage(input, stream);
-    all_reduce(output, stream);
+    all_reduce(output, stream, /*allow_owner_path=*/false);
   }
 
   void all_gather_fused(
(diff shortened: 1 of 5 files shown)


Flink-ddd (approved):
LGTM

frank-2077 (approved):

PR #375, opened 2026-09-01
feat: enable Triton kernels on MUSA

Arlo-mt:
## Summary

Thank you for merging the previous PR that added the MUSA backend support. Building on that foundation, this PR adds MUSA support for Triton-backed RL-Kernel operators.

MUSA tensors are not reported as CUDA tensors by PyTorch, so the existing
`.is_cuda` and `torch.cuda.is_available()` checks could incorrectly reject
MUSA execution or select the CPU fallback. This change adds explicit MUSA
device handling while preserving the existing CUDA and ROCm paths.

## Changes

- Add MUSA device detection through `torch_musa` and `torch.musa`.
- Add the MUSA device and backend constants.
- Update Triton operator validation to accept MUSA tensors.
- Enable MUSA Triton dispatch for:
  - `logp`
  - `grpo_loss`
  - `ratio_kl`
  - `linear_logp`
  - `det_gemm`
  - `batch_invariant_logp`
  - `rms_norm`
  - `embedding`
  - `silu`
  - `swiglu`
  - `rope`
- Keep PyTorch fallback paths available when a Triton backend is unavailable.
- Add MUSA-specific dispatch and `linear_logp` forward/backward coverage.
- Update accuracy and dispatch tests to distinguish MUSA from CPU and CUDA.

This PR does not add MUSA native C++/MUSA kernels. Native MUSA kernel support
is handled separately in the `MUSA-support-native-kernels` branch.

## Validation

Validated locally on:

- GPU: Moore Threads S5000
- PyTorch: `2.9.0.post1+musa5.1.2mp22`
- Triton: `3.2.0`
- MUSA architecture: `mp_22`

Results:

- MUSA Triton operator smoke tests: `5/5 passed`
- MUSA dispatch and accuracy tests: `7/7 passed`
- CPU-safe and dispatch unit tests: `100 passed`
- Python compilation and whitespace checks passed

## Compatibility

- CUDA and ROCm-specific dispatch paths are unchanged.
- CUDA-only tests such as SM90, NCCL, and CUDA extension tests remain
  CUDA-specific.

## Summary by CodeRabbit

- **New Features**
  - Added MUSA support across embedding, loss functions, normalization, rotary embeddings, and related GPU operations.
  - Added deterministic linear log-probability calculations, including tensor-parallel support.
  - MUSA dispatch now selects validated optimized backends with PyTorch fallbacks where needed.
  - Added support for contract-aware attention selection on MUSA.

- **Bug Fixes**
[...]

PR #204, opened 2026-07-02
[WS1][kernels] Batch-invariant logprob (CUDA)

KJLdefeated:
# [WS1][kernels] SM90 Batch-Invariant logprob CUDA Kernel

## Summary

Adds a Hopper (SM90) TMA CUDA kernel for the batch-invariant selected-token log-probability operator introduced in #199  plus a Native/Triton/CUDA benchmark. The operator computes, from **already materialized** logits:
```text
logp[t] = logits[t, target_ids[t]] - logsumexp(logits[t, :])
```

The kernel streams the vocab through shared memory with TMA bulk-tensor copies and folds it into a running online log-sum-exp, so it never materializes the `[N, V]` softmax and its extra forward memory is ~0. It is registered as the
top-priority CUDA backend on Hopper and falls back cleanly everywhere else.

| Path | Status |
| --- | --- |
| CUDA SM90 forward (`csrc/cuda/batch_invariant_logp_kernel_sm90.cu`) | TMA online-softmax, bf16 + fp32, one CTA per row. New. |
| Python op (`BatchInvariantLogpSM90Op`) | Autograd wrapper; TMA forward + tile-wise backward. New. |
| Backward | Reuses #199's Triton tile-wise backward via the forward-saved `lse` (no `[N, V]` materialization). |
| Registry | Hardware-gated: inserted at the front of the CUDA `batch_invariant_logp` list only when the SM90 symbol is compiled on `cc_major == 9`. |
| Fallback | Non-bf16/fp32, unaligned vocab row stride, or non-Hopper → Triton, else native. |
| Benchmark | `benchmarks/benchmark_batch_invariant_logp.py` compares Native/Triton/CUDA. New. |

This builds on the open branch for #199 (Native + Triton). Reviewed after #199 merges.

## Implementation

- **New kernel** `csrc/cuda/batch_invariant_logp_kernel_sm90.cu`.
- **Binding** in `csrc/ops.cpp` and **build source** in `setup.py`'s `sm90_srcs`; stub in `rl_engine/_C.pyi`.
- **Python op** `rl_engine/kernels/ops/cuda/loss/batch_invariant_logp.py`:
  - `_BatchInvariantLogpSM90Function` forward calls the compiled kernel;
  - backward reuses Triton `_batch_invariant_logp_bwd_kernel`.
- **Registry** `rl_engine/kernels/registry.py`.
- **Docs/benchmark**: updated `docs/operators/batch-invariant-logp.md`; added the benchmark script.

## Validation environment

[...]

PR #376, opened 2026-09-01
Musa support native kernels

Arlo-mt:


zhangj1an (changes requested):
Thanks! It is already almost there, after addressing the following we can merge it.

 Please refer to PR 204 (https://github.com/RL-Align/RL-Kernel/pull/204) as an example when adding new kernels.
1. Test breakage: tests/test_rms_norm.py calls torch.musa.is_available() unguarded and fails with AttributeError on non-MUSA platforms (verified). Use device_ctx.is_musa or a hasattr guard.
 2. Please register musa kernels in rl_engine/kernels/gtest/operator_specs.py ("musa" candidates), docs/operators/*.md Backends rows, rl_engine/_C.pyi stubs for the new symbols, and a benchmark script.
3. Bitwise evidence: for copy/elementwise ops (embedding/rope/silu/swiglu) add torch.equal bitwise tests vs the same-dtype PyTorch reference (fp32 internally, single cast at the end); for reductions (logp/lm_head/rmsnorm/attention/det_gemm) add bitwise batch-invariance tests + contract-tolerance accuracy tests, mirroring tests/test_batch_invariant_logp.py.
 4. Please rebase your branch on the test branch and change the PR base to test. Also update TestMusaPlatform expectations (the native-first musa map conflicts with the current assertions on both main and test), and coordinate the musa priority map with #375 (decide merge order / final native-first + triton-fallback map).
 5. Fix DCO sign-off and the unreachable device_ctx.is_musa branch in tests/test_logp.py.

Arlo-mt:
Thanks！I will submit PRs according to PR 204 (https://github.com/RL-Align/RL-Kernel/pull/204) format.

PR #378, opened 2026-09-02
feat(ascend): add RoPE kernel

erfgss:
## Summary

This PR adds an Ascend C backend for Rotary Position Embedding (RoPE).

- Add an Ascend C RoPE kernel using the GPT-NeoX/Hugging Face rotate-half convention.
- Support FP16, BF16, and FP32 inputs with FP32 intermediate computation.
- Support position layouts in both `[S]` and `[B, S]` formats.
- Add autograd support by applying the inverse rotation during backward.
- Register `RoPEAscendOp` as the preferred NPU backend, with the native PyTorch implementation as a fallback.
- Integrate the Ascend source files into the `_C_npu` build process.
- Consolidate Ascend-specific tests into `tests/test_rope.py`.
- Update the operator documentation and build instructions.

## Implementation

The Ascend C kernel applies RoPE using the following rotation:

```text
out[i]       = x[i]       * cos - x[i + D/2] * sin
out[i + D/2] = x[i + D/2] * cos + x[i]       * sin
```

The Python wrapper generates FP32 cosine and sine tables and converts the input into the row layout expected by the kernel. Per-batch position IDs are supported by rearranging `[B, H, S, D]` inputs before launching the kernel and restoring the original layout afterward.

The backward pass reuses the same Ascend C primitive with a negative sine sign, which applies the transpose of the forward rotation.

The shared `_C_npu` module registration lives in `csrc/ascend/npu_module.cpp`, the single consolidated pybind entry shared by Ascend branches. It was renamed from `ops_ascend.cpp` per review feedback so that merging other Ascend branches (e.g. `erfgss/rmsnorm_ascend`) does not produce duplicate `PYBIND11_MODULE` definitions with colliding `PyInit__C_npu` symbols. Both the existing batch-invariant log-probability operator and the new RoPE operator are exported from the same extension.

## Validation

The following tests were run:

```bash
python -m pytest \
  tests/test_rope.py \
  rl_engine/tests/test_dispatch.py \
  tests/test_operator_inputs.py \
  tests/test_envs.py \
  -q
```

Result:

```text
71 passed, 9 skipped
```

The skipped tests require an Ascend NPU and an extension built with CANN/BiSheng. Ascend C compilation and on-device validation were not available in the local environment.

## Summary by CodeRabbit

* **New Features**
  * Added an Ascend NPU RoPE backend supporting FP16, BF16, and FP32 tensors, configurable positions, gradients, and empty inputs.
  * NPU RoPE dispatch now prioritizes the Ascend implementation with a native PyTorch fallback.
  * Added Ascend C extension build and installation support.

* **Documentation**
  * Updated RoPE documentation and Quick Start instructions for Ascend support.

* **Bug Fixes**
  * Resolved duplicate binding behavior for the Ascend log-probability operation.

### DIFF
--- a/csrc/ascend/npu_module.cpp
+++ b/csrc/ascend/npu_module.cpp
@@ -1,0 +1,25 @@
+// SPDX-License-Identifier: Apache-2.0
+// Copyright (c) 2026 RL-Kernel Contributors
+
+#include <torch/extension.h>
+
+#include <vector>
+
+std::vector<torch::Tensor> batch_invariant_logp_ascend_forward(torch::Tensor logits,
+                                                               torch::Tensor target,
+                                                               int64_t ignore_index);
+
+torch::Tensor rope_apply_ascend_forward(torch::Tensor x,
+                                        torch::Tensor cos,
+                                        torch::Tensor sin,
+                                        double sin_sign);
+
+PYBIND11_MODULE(TORCH_EXTENSION_NAME, m)
+{
+    m.def("batch_invariant_logp_ascend",
+          &batch_invariant_logp_ascend_forward,
+          "Batch-invariant selected-token log-probability (Ascend C forward)");
+    m.def("rope_apply_ascend",
+          &rope_apply_ascend_forward,
+          "GPT-NeoX/HF rotate-half RoPE apply (Ascend C forward/backward primitive)");
+}
--- a/README.md
+++ b/README.md
@@ -123,6 +123,10 @@ python -m pip install -e .
 # Native CUDA or ROCm extension (install a matching PyTorch build first)
 RL_KERNEL_REQUIRE_EXT=1 python -m pip install --no-build-isolation -e .
 python -c "import rl_engine._C as _C; assert hasattr(_C, 'fused_logp'); print(_C.__file__)"
+
+# Ascend C extension (Linux host with matching CANN + torch_npu)
+KERNEL_ALIGN_FORCE_ASCEND=1 python -m pip install --no-build-isolation -e .
+python -c "import rl_engine._C_npu as C; assert hasattr(C, 'rope_apply_ascend'); print(C.__file__)"
 ```
 
 ### Contributions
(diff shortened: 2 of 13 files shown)


csrc/ascend/npu_module.cpp
```diff
@@ -0,0 +1,25 @@
...
+std::vector<torch::Tensor> batch_invariant_logp_ascend_forward(torch::Tensor logits,
+                                                               torch::Tensor target,
+                                                               int64_t ignore_index);
+
+torch::Tensor rope_apply_ascend_forward(torch::Tensor x,
+                                        torch::Tensor cos,
+                                        torch::Tensor sin,
+                                        double sin_sign);
+
+PYBIND11_MODULE(TORCH_EXTENSION_NAME, m)
```
zhangj1an, line 17: Nit:  Could you rename ops_ascend.cpp → npu_module.cpp? This is because Ascend branches (including erfgss/rmsnorm_ascend) use csrc/ascend/npu_module.cpp as the single consolidated pybind entry. If we use current naming, the two PYBIND11_MODULE definitions will collide (duplicate PyInit__C_npu) when branches merge.

zhangj1an (approved):
LGTM. I have checked this branch (all three dtypes with `--check-grad`):
```
  python scripts/check_operator.py --op rope --candidate ascend --device npu
      --dtype {fp32,bf16,fp16} --batch 2 --seq 16 --vocab 257 --normalized-dim 4096
      --check-grad
```
  - fp32: forward and gradient are **bitwise identical** to the PyTorch reference (max_abs = 0.0).
  - bf16: forward max_abs 1.53e-2 within the (2e-2, 1.6e-2) contract tolerance; gradient bitwise (0.0). The forward diff vs the fp32
  gold is the output cast quantization, not kernel error — the kernel is bitwise identical to the PyTorch dtype-path reference for
  bf16 as well.
  - fp16: forward within (1e-3, 1e-3) (max_rel 7.16e-4); gradient bitwise (0.0).

zhangj1an:
Thanks, LGTM

inaniloquentee (approved):

PR #383, opened 2026-09-03
[DSv4][P1-S0] Start kit for the P1 work package (mHC + RMSNorm)

zhangj1an:
## What

Start kit for DSV4 Project 1: mHC + RMSNorm deterministic forward/backward.

This PR provides the API, numerical contract, PyTorch golden reference, fixtures, provider interface, and acceptance tests for P1. It intentionally contains no CUDA or Triton kernels.

If you claim a P1 task:

1. use `oracle.py` as the golden numerical reference;
2. implement the corresponding CUDA and/or Triton kernel;
3. expose it through `MHCProvider`;
4. make `scripts/check_p1.py` pass byte-for-byte.

PR #204 is the reference submission pattern: kernel + binding + registry gating + clean fallback + tests + benchmark + validation environment.

## Tasks

| Task | Scope | Status | Person in charge | PR number | 
| `P1-1` | `hc_split_sinkhorn` — controller mapping + Sinkhorn, fwd/bwd | 🙋 | | |
| `P1-2` | `fp32_gemm_rms` — FP32 controller projection + RMS, fwd/bwd | ⏳ | | |
| `P1-3` | `mhc_post` — write sublayer output back to four residual streams, fwd/bwd | ⏳ | | |
| `P1-4` | `mhc_pre` / `h_aggregate` — four-stream aggregation, fwd/composite bwd | ⏳ | @nodeeeeee| #366|
| `P1-5` | `rmsnorm_residual` — RMSNorm + residual fork, fwd/bwd | ⏳ | @why1te | |
| `P1-6` | `fp32_gemm_rms` TP/SP contract | ⏳ | | |
| `P1-7` | mHC rank-local invariance under TP/SP/CP/DP/PP | 🙋 | | |
| `P1-8` | `rmsnorm_residual` TP/SP semantics + cross-rank `dgamma` | ⏳ | | |

WS1 tasks (`P1-1`–`P1-5`) are independent and can be implemented in parallel from this start kit.

`P1-4` may call the oracle versions of `hc_split_sinkhorn_bwd` and `fp32_gemm_rms_bwd`, so it does not need to wait for `P1-1` or `P1-2`.

WS2 (`P1-6`–`P1-8`) attaches through `LayerContract.placement` and `check_capability`.

## What's inside

| Path | Purpose |
| --- | --- |
| `rl_engine/mhc/reduction.py` | Pinned reduction trees |
| `rl_engine/mhc/contract.py` | P1 contracts and fingerprints |
| `rl_engine/mhc/oracle.py` | FP32 golden reference for all P1 fwd/bwd operators |
| `rl_engine/mhc/provider.py` | `MHCProvider`, reference provider, stub, capability gate |
| `rl_engine/mhc/fixtures.py` | Seeded fixtures + golden hashes |
| `rl_engine/mhc/trace.py` | Boundary hashes + first-divergence reporting |
| `scripts/check_p1.py` | Backend acceptance command |
| `tests/test_p1_*.py` | CPU contract tests |

## Numerical contract

CUDA/Triton implementations must reproduce the arithmetic and reduction order in `oracle.py`.

Only two reduction trees are allowed:

- Long reductions: FP32 left fold in ascending index order.
- 4-element reductions: `(a0 + a1) + (a2 + a3)`.

The following are forbidden where they change reduction order:

- Split-K / Stream-K
- atomic partial accumulation
- runtime-dependent reduction trees

The golden reference does not use `torch.sum`, `matmul`, `mean`, `einsum`, or `norm` for contracted reductions.

### 1. Read the oracle first

Start from the corresponding forward/backward implementation in `oracle.py`.

The oracle defines:

- arithmetic order;
- reduction order;
- dtype boundaries;
- downcast points;
- expected output bytes.

Your CUDA/Triton implementation should reproduce those semantics.

### 2. Add it to a provider

```python
from rl_engine.mhc.provider import ReferenceProvider


class MyCudaProvider(ReferenceProvider):
    name = "my-cuda"
    numeric_profile = "cuda-ffma-strict-v1"

    def mhc_post_fwd(self, r_old, y, c, post):
        return my_cuda_kernel(r_old, y, c, post)
````

Override only the operator your task implements. Everything else may continue using the golden reference.

`numeric_profile` is part of the numerical contract. Do not claim strict bit equivalence unless the implementation actually provides it.

### 3. Run acceptance

```bash
python scripts/check_p1.py \
    --provider my_backend.p1_provider:MyCudaProvider \
    --device cuda
```

Every checked boundary must match the oracle according to the declared numerical profile.

The checker also verifies:

> same row, different batch / padding / stride ⇒ identical bytes

On failure it reports the first diverging boundary.

## Kernel PR checklist

Follow PR #204.

* [ ] CUDA and/or Triton kernel
* [ ] Binding + build integration
* [ ] Python/autograd wrapper + `MHCProvider`
* [ ] Hardware-gated registry entry with clean fallback
* [ ] `check_p1.py` passing result
* [ ] Forward/backward test sweep
* [ ] Bitwise batch-invariance test
* [ ] Benchmark: latency + peak extra memory
* [ ] Validation environment table
* [ ] Operator documentation

## Contract changes

Do not modify the oracle or golden hashes just to make a kernel pass.

If the numerical contract itself needs to change, get maintainer approval first, then regenerate the manifest:

```bash
python -m rl_engine.mhc.fixtures --write-manifest
```

## Not in this PR

This start kit does not contain CUDA/Triton kernels or framework injection.

RoPE, attention, MoE, and later TP/SP/CP/DP/PP integration are handled separately.

The CPU fixtures use reduced geometry to keep the serial oracle cheap. Production geometry remains pinned by `LayerContract.assert_production()`:

```text
hidden = 4096
K      = 16384
N      = 24
```

## Current status

* 54/54 CPU tests pass
* Hand-written backward paths are cross-checked against autograd
* Reference provider passes `check_p1.py`
* Stub provider fails closed as intended
* ruff / flake8 / black / isort / mypy clean


inaniloquentee (approved):

ZeroKernel798:
Hi @zhangj1an, is P1-3 (`mhc_post`) still open? I'd like to take it if nobody is on it.

I've been through the P1 start kit — the `mhc_post_fwd` / `mhc_post_bwd` reference
(tagged `P1-D3`) at `oracle.py:394-448`, the two pinned trees in `reduction.py`
(`fixed_sum` ascending fold, and `stream4_sum`'s `(a0+a1)+(a2+a3)`), and the design
doc's D1–D10 decisions.

Plan: implement CUDA (then Triton) via `MHCProvider`, override only `mhc_post_fwd` /
`mhc_post_bwd`, and get `scripts/check_p1.py --device cuda` byte-equal, following the
submission pattern of #204 / #419.

Environment: 8x H20 (SM90), CUDA 13, TORCH_CUDA_ARCH_LIST=9.0+PTX.

If it's already assigned internally, no problem — just let me know.

Issue #64, opened 2026-06-02
[FEAT][alignment]: audit GRPO loss reduction semantics and add production-scale validation


inaniloquentee:
## Goal

Make GRPO loss reduction semantics explicit and validate them under ragged,
large-shape, production-like rollout workloads.

## Background

Issue #16 averages over active completion tokens globally. This is correct for
the current Kernel-Align objective contract, but DeepSeek-style GRPO may require
sequence-level averaging before group/batch aggregation depending on the target
formula. The behavior should be explicit, configurable if needed, and tested on
ragged completions.

## Tasks

- Document the target GRPO loss formula.
- Compare global-token mean vs per-sequence mean behavior.
- Add config support if both reduction modes are required.
- Add tests where completion lengths differ and the two reductions produce
  different results.
- Add larger CUDA benchmark profiles for realistic group sizes, completion
  lengths, and vocabulary sizes.
- Record memory/time/drift evidence in issue docs.
- Keep smoke tests fast and deterministic.

## Acceptance Criteria

- Loss reduction mode is explicit in config and metrics.
- Ragged completion tests prove the selected behavior.
- CUDA benchmark evidence covers both small smoke and larger production-shaped inputs.

inaniloquentee:
This is the PR: https://github.com/Flink-ddd/RL-Kernel/pull/67

Issue #63, opened 2026-06-02
[FEAT][alignment]: use real policy logprobs and reward providers for production GRPO


inaniloquentee:
## Goal

Move GRPO training from smoke-style local inputs to production-grade rollout inputs:
real old-policy logprobs, real reference-policy logprobs, and pluggable reward sources.

## Background

Issue #16 implements the GRPO objective and worker integration, but the current
training-worker helper still synthesizes old/ref logps from current logps for
local smoke paths. Rewards can be consumed from payloads and strict validation
exists, but reward computation itself is not yet modeled as a production adapter.

## Tasks

- Define the rollout/training batch contract for:
  - current selected logprobs;
  - old-policy selected logprobs captured during rollout;
  - reference-policy selected logprobs;
  - scalar sequence rewards;
  - prompt group ids and completion masks.
- Extend rollout payload parsing to consume old/ref logps when supplied.
- Add shape/device/dtype/mask validation for logprob tensors.
- Add a reward provider protocol for:
  - payload scalar rewards;
  - rule-based verifier callback;
  - optional model-backed reward callable.
- Keep local smoke fallback clearly labeled and disabled in strict production mode.
- Add tests where current, old, and reference logps are distinct tensors.
- Add tests for missing/mismatched logps and rewards.

## Acceptance Criteria

- GRPO worker loss uses real supplied old/ref logps when available.
- Production strict mode rejects incomplete rollout/reward/logprob payloads.
- Reward computation is pluggable without importing vLLM, DeepSpeed, or Ray at module import time.

inaniloquentee:
This is the PR: https://github.com/Flink-ddd/RL-Kernel/pull/67

PR #387, opened 2026-09-04
[DSv4][P5-5] Shared Expert MLP: strict CUDA + Triton kernels

jyizheng:
Implements `shared_expert_mlp_fwd/bwd` (#64) per the `P5-S0` contract: every valid token runs fc1 -> one-round SwiGLU -> fc2 on BF16 frozen weights; backward returns dX only (FP32 accumulator), and the shared output stays independent of the routed path (combine belongs to P6).

Both backends reproduce `oracle-fp32-serial-v1` byte-for-byte on the same device: one lane owns one output element and reduces serially in ascending k, multiply and add rounded separately (`__fmul_rn`/`__fadd_rn` on CUDA, libdevice `mul_rn`/`add_rn` on Triton). No cross-lane floating-point reduction exists, so results are batch/padding invariant by construction. The one-round SwiGLU core ships in shared mode (`p_s = None`, no clamp, per S0 decision D6) as the reuse point for P5-2 (#63).

## Acceptance (B300 / sm_103, torch 2.14+cu130)

```
check_p5.py --provider ...:CudaSharedExpertProvider   --device cuda  ->  PASS (all boundaries byte-equal)
check_p5.py --provider ...:TritonSharedExpertProvider --device cuda  ->  PASS (all boundaries byte-equal)
pytest tests/test_shared_expert_mlp.py                               ->  14 passed
```

Tests cover: fixture byte-equality (`shared_t1`/`shared_t16`), batch/padding invariance (`fwd(x)[t] == fwd(x[t:t+1])` per row), no `dW` (weight hashes stable, no grad), shared/routed boundary independence, CUDA/Triton cross-backend byte-equality at (T=256, H=1024, F=512), and fail-closed behavior (CPU input / missing backend / foreign numeric profile raise; no silent oracle fallback).

## Benchmark (torch-native vs Triton vs CUDA, H=4096 F=2048, fwd+bwd ms)

| T | torch-native (cuBLAS, non-deterministic) | Triton strict | CUDA strict |
|---|---|---|---|
| 16 | 0.212 | 2.08 | 2.05 |
| 256 | 0.193 | 24.4 | 25.2 |
| 2048 | 0.233 | 191 | 199 |

The strict backends are asserted byte-equal to each other on every shape. The gap vs cuBLAS is the cost of the serial one-lane-per-element reduction; epilogue/perf work is the P5 step-10 follow-up, out of scope here.

## Notes

- Sigmoid is transcendental, so its bits follow the libm implementation: nvcc `expf` bit-matches `torch.sigmoid` (0/4M mismatches on a device probe) while `tl.exp` (exp2-based) and libdevice `__nv_expf` do not (~45% / ~10% of values off by 1 ulp). The CUDA kernel computes `1/(1+expf(-x))` inline; the Triton path takes `torch.sigmoid(gate)` as a kernel input and fuses the remaining SwiGLU math.
- All Triton arithmetic goes through libdevice `*_rn`: the compiler may contract `a*b+c` into an FMA, which showed up as a 1-ulp dgate drift on 2/262144 elements at T=256. A cross-backend byte-equality test at that size locks the regression in.
- `provenance()` records `split_k=1`, `reduction=serial-ascending-k`, no-FMA rounding, per the P5-5 provenance requirement.


KJLdefeated (commented):
Thanks for the clean and clear implementations. I pulled the branch onto `dsv4-p5-dev` and verified both backends on an H100 (torch 2.14+cu130, Triton 3.8). CUDA is byte-equal to the oracle on both fixtures, the T=256 cross-backend check, and a 2M-element SwiGLU sweep, and backward is batch-invariant too. Good Work!

Some issues about performance:
The performance is 1000x slower than cublas. I think there is still room for optimize.
Can you reference to [Det GEMM](https://github.com/RL-Align/RL-Kernel/blob/main/csrc/cuda/gemm/det_gemm_kernel.cu) to see if we can reuse this kernel or reference this design?
Note that reusing it may not align with oracle I provided, but we can ignore it for now.
Our goal is ensure batch-invariance and also high performance.

jyizheng:
Thanks for verifying on H100! Agreed on the perf gap — the serial one-lane-per-element path was correctness-first. I'll add a performance path on top of `det_gemm_kernel.cu` (batch-invariant tiled GEMM) registered as a separate numeric profile, keeping the strict path as the oracle-parity gate, and post benchmark numbers here. The SwiGLU core stays as is since the cost is entirely in the GEMMs. Also signed off the commits for DCO.

jyizheng:
Performance path is in (last two commits):

- `shared-expert-cuda-det` (`p5-det-gemm-v1`): reuses `det_gemm` through two new FP32-output wrappers (`det_gemm_fwd_out_fp32` / `det_gemm_fwd_rhs_transposed_out_fp32` — they just expose the existing `gemm_dispatch` `output_fp32` path), so the P5-5 round positions are preserved and only the reduction order changes. Built with `KERNEL_ALIGN_DET_GEMM_SM90=1`; the TMA+mma.sync path compiles and runs on sm_103a as well as sm_90a.
- `shared-expert-triton-det` (`p5-triton-dot-v1`): `tl.dot` with fixed 64x64x32 tiles, ascending-k, no autotune, no split-K.

Both are selected explicitly by provider name — the strict providers stay the oracle-parity gate and are never substituted.

Benchmark (B300/sm_103, H=4096 F=2048, fwd+bwd ms):

| T | torch-native | cuda strict | triton strict | cuda-det | triton-det |
|---|---|---|---|---|---|
| 16 | 0.25 | 2.05 | 2.08 | 0.79 (3.2x native) | 0.16 (0.6x) |
| 256 | 0.41 | 25.2 | 24.4 | 0.75 (1.9x) | 0.19 (0.5x) |
| 2048 | 0.35 | 199 | 191 | 2.30 (6.7x) | 0.52 (1.5x) |

So the gap vs cuBLAS goes from ~100-850x (strict) to ~0.5-7x, and `triton-det` is at or below cuBLAS latency on these shapes while staying deterministic and batch-invariant (new tests pin repeat-run byte-equality and per-row invariance for both perf profiles).

One observation from wiring `det_gemm` in: its BK=32 partials are rounded to BF16 and the K dimension is merged with the BF16 mid-split tree, so its deviation from the FP32-serial oracle is BF16-tree-sized (maxabs ~4e-2 on K=512 vs ~8e-3 for the full-FP32 `tl.dot` path). That is presumably the intended TP-equivalence tradeoff — the closeness tests encode per-profile tolerances accordingly. If you'd rather have an FP32-tree variant of `det_gemm` for training-side use, happy to follow up.

KJLdefeated (approved):
LGTM

PR #407, opened 2026-09-12
docs(ws1): add numerical precision standard with threshold provenance

maxiaosong1124:
## Summary

Adds `docs/design/ws1-numerical-precision-standard.md`, a single citable document defining the WS1 ground truth, judgment methods and threshold provenance. Documentation only; no code or contract changes.

The document keeps normative targets and current implementation status strictly separate. Sections 1-8 are the standard; section 9 is an audit of where the repository currently diverges from it; section 11 is the plan to close those gaps.

Status is **Draft Normative**. Until the seven blocking items in section 11 are closed, the repository should not be described as fully satisfying this standard.

## Normative content (sections 1-8)

- **FP32 reference definition** (§3.1): the candidate and the reference consume the same low-precision input values; the reference upcasts them to FP32, evaluates the operator's original definition in FP32 with TF32 and autocast disabled, and keeps the output in FP32. This places input quantization error outside the measurement.
- **Why FP32 is required, not preferred** (§3.3): the softmax / log-sum-exp forward error bound stops constraining anything once `n ≳ 1/u`. With `u_bf16 = 2^-8` that boundary is `n ≈ 256`, while Qwen3-8B reduces over 4096 (RMSNorm) and 151936 (vocabulary). A BF16 implementation cannot serve as the baseline because its own error guarantee is too weak to adjudicate the quantity under test.
- **Limits of reference authority** (§3.4): authority comes from the dtype path, not the library name. The cuBLAS reproducibility preconditions are used to show why an unpinned cuBLAS path cannot be the bitwise oracle.
- **Threshold provenance** (§6): every row is traced to PyTorch, NVIDIA TransformerEngine or H20 measurement, and classified as direct adoption, partial reference, or measured calibration. A match on `atol` or `rtol` alone is not reported as a full match. `clip_interval` is stated as a PPO policy interval rather than a floating-point tolerance.
- **Error-magnitude placement** (§7): BF16 thresholds sit at 4u-5u; the FP32 reduction threshold `1e-4` sits between the empirical scale and the sequential worst-case bound `gamma_n ≈ 2.44e-4`, with the caveat that real kernels use different reduction topologies.

## Implementation audit at `6ceeb62` (section 9)

Six gaps are recorded, each tied to a source location verified at that commit.

1. **Deterministic GEMM accumulation** (§9.2). Both the CUDA and Triton paths reduce K with a mid-split tree whose leaves accumulate in FP32 and whose merges round to BF16. The kernel header documents this and gives the rationale: a contiguous half-K split makes a simulated TP=2 reproduce TP=1. This is a deliberate design for TP-invariance, not a defect. The gap is documentary: the contract declares `accumulation_dtype: float32` globally and requires per-operator exceptions to be declared, but carries no declaration for `det_gemm`. Section 9.2 states the two available resolutions and section 11 makes choosing one the first blocking item.
2. **`det_gemm` FP32 reference** (§9.3). The gold path resolves to `torch.matmul` and returns BF16 for BF16 inputs, so the accuracy judgment compares two BF16 GEMMs. Independent of item 1: this concerns the reference, that one concerns the candidate.
3. **Invariance comparator** (§9.4). Both comparators upcast to FP32 and call `torch.allclose`. Since the upcast is lossless this equals exact numerical equality, with the gap confined to `±0` and NaN handling. Historical C3/C4/C8 results should be described as zero-tolerance numeric equality rather than bitwise verification.
4. **CUDA generic logp output dtype** (§9.5). `fused_logp_forward` allocates from `logits.options()`, so BF16 logits yield BF16 logprobs against the FP32 aggregate policy. An FP32 entry point already exists in the same file.
5. **RMSNorm rounding order** (§9.6). The reference uses kernel order while the contract pins neither order. The two differ by roughly 1 ulp and the choice affects the achievable chain-parity floor.
6. **Legacy tolerances** (§2.2). 306 hard-coded `atol=` occurrences across 36 test files bypass the contract.

TF32 claims are scoped to the required path, which is clean: `det_gemm` sets `allow_tf32=False`, `linear_logp` sets `input_precision="ieee"`, and `standard_attn` uses no `tl.dot`. `triton_attn.py` and `chunked_flash_attn.py` are outside the C8 matrix, the latter reached only through the ROCm path.

## Evidence scope (section 10)

The checked-in C8 H20 evidence is bounded to its own source commit and environment. It does not attest to current `main` and does not replace final-commit GPU CI.

## Notes for review

- External citations use commit permalinks that were fetched and verified, so the quoted text is stable.
- No threshold values are changed by this PR. Proposed changes (tightening `approx_kl0`, adding `mean_abs_dlogp`, recalibrating `max_abs_dlogp`) are listed in §11.2 as requiring a new contract version, suggested `ws1-c1-v3`.

### DIFF
--- a/docs/design/ws1-numerical-precision-standard.md
+++ b/docs/design/ws1-numerical-precision-standard.md
@@ -1,0 +1,576 @@
+# WS1 Numerical Precision Standard: Ground Truth Definition, Judgment Methods, and Threshold Provenance
+
+| Field | Value |
+|---|---|
+| Document status | **Draft Normative** |
+| Document version | v1.0 |
+| Document date | 2026-09-11 |
+| Precision contract | `rl_engine/kernels/gtest/tolerance_contract.json` |
+| Contract version | `ws1-c1-v2` |
+| Implementation audit baseline | `6ceeb62e073d74e00a8c5b56af538ee1a7417773` (`origin/main`) |
+| Target model | Qwen3-8B Dense (`ws1-qwen3-8b-dense-primary-v6`) |
+| Required backend profiles | `cuda_bf16`, `triton_cuda_bf16` |
+| Optional precision | FP16 (complete when declared; not required for WS1 exit) |
+| Out of WS1 scope | FP8, ROCm, Ascend, and any optional fused path not listed in the C8 required-op matrix |
+
+> **On document status.** The normative sections (§1–§8) define the numerical precision targets and judgment methods for WS1 and may be cited as the project's standard. Section 9 records where the current repository diverges from that standard, Section 10 bounds the applicability of historical evidence, and Section 11 gives the plan for closing the gaps. Until every blocking item in Section 11 is closed, this repository must not be described as fully satisfying this standard. Normative targets and implementation status are stated separately throughout.
+
+---
+
+## 1. Summary
+
+WS1 separates numerical judgment into four classes, so that standards of different natures are not collapsed into a single "tolerance table":
+
+| Judgment | Compared quantities | Judgment mode |
+|---|---|---|
+| Forward accuracy | BF16 candidate ↔ FP32 reference | Element-wise tolerance |
+| Gradient accuracy | BF16 backward ↔ FP32 reference gradient | Element-wise tolerance |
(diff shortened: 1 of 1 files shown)


Flink-ddd (approved):
LGTM

PR #400, opened 2026-09-10
docs(rocm): publish G10/G11 200-step RL results

inaniloquentee:
## What this PR adds

- Publishes the completed ROCm Qwen3-8B 200-step G10 production P/P and G11 RL-Kernel R/R consistency experiment derived from PR #396.
- Adds the complete paired-step CSV, structured summary, strict-validation summary, and two reproducible consistency figures.
- Adds local-log regeneration and W&B upload scripts so the complete RL history and source artifacts can be reproduced and published without rerunning training.

## Completed experiment scope

| Group | Attention / FFN / logp route | Ray submission | Result |
|---|---|---|---|
| G10 | production / production (P/P) | `raysubmit_Rj3ZWxq3mphM2bFL` | SUCCEEDED, 200/200 |
| G11 | RL-Kernel / RL-Kernel (R/R), strict consistency | `raysubmit_twpgQ7pxvniaXWXC` | SUCCEEDED, 200/200 |

Both runs used the same frozen workload and source configuration. Only the intended G10 P/P versus G11 R/R implementation routes and identity-derived paths differ.

| Item | Matched value |
|---|---|
| Hardware | 1 node, 8x AMD Instinct MI300X 192GB |
| Actor parallelism | TP4 / CP2 / PP1 |
| Rollout engines | 2 engines, TP4 each |
| Horizon | 200 rollout/training steps |
| Seeds | training `1234`, rollout `1234` |
| Sampling | 1 prompt x 8 samples = global batch 8 |
| Response limit | 7,168 tokens |
| Dynamic batching | maximum 4,096 tokens/GPU |
| vLLM memory utilization | 0.38 |
| Rollout log probabilities | enabled |
| Reference model / KL loss | enabled, coefficient `0.001` |
| Torch profiler | disabled |

## Version provenance

| Arm | RL-Kernel | VIME | Megatron-LM | Notes |
|---|---|---|---|---|
| G10 | [`7a9f3b5`](https://github.com/RL-Align/RL-Kernel/commit/7a9f3b5657ac380ae2419539975498513d18c490) | `c80200e` | `1dcf0da` | Native P/P route |
| G11 | [`7a9f3b5`](https://github.com/RL-Align/RL-Kernel/commit/7a9f3b5657ac380ae2419539975498513d18c490) | `c80200e` | `1dcf0da` | Strict RL-Kernel R/R route |
[...]

PR #423, opened 2026-09-17
fix(vime): avoid double temperature scaling for reused logits

inaniloquentee:
## Summary

- avoid applying rollout temperature twice for legacy VIME requests that reuse pre-scaled local logits
- preserve one temperature application for current VIME requests with unscaled local logits
- keep hidden-state recomputation on the original request temperature
- add numerical regression coverage for both real VIME contracts at temperature 0.7
- change RL-Kernel only; no VIME source changes

## Contract details

The PR #400 experiment pins VIME commit `c80200e`. That revision divides `request.logits` by `rollout_temperature` before provider dispatch and sets `metadata["logits_are_temperature_scaled"] = True`. Reapplying `request.temperature` in RL-Kernel therefore turns 0.7 into an effective 0.49.

VIME PRs #423 and #424 use the newer contract: provider-bound logits remain unscaled, the metadata marker is absent, and `request.temperature` must still be applied once. This change supports both contracts. The marker is consumed only by the reused-local-logits path; hidden-state recomputation continues to use the request temperature.

## Validation

- `python -m pytest -q tests/test_vime_linear_logp_provider.py`: 10 passed
- Black, isort, Flake8, and `git diff --check` passed for the changed files
- patched PR #400 G11 strict run: first 4,165 tokens had mismatch count 0 and max absolute logp diff 0.0


Flink-ddd (approved):
LGTM

maxiaosong1124 (approved):
LGTM!

PR #418, opened 2026-09-17
docs(blog): RL-Kernel × vime × AMD train-rollout bitwise consistency

maxiaosong1124:
## Summary

- Add an English blog article on train-rollout bitwise consistency with RL-Kernel, vime, AMD, and Qwen3-8B.
- Cover numerical execution contracts, operator ablations, and CUDA/ROCm 200-step results.
- This draft includes only the English blog document.

## Validation

- The local preview build passed `mkdocs build --strict -f mkdocs.yaml`, with Git timestamp fallback warnings for uncommitted documents.
- The English page returned HTTP 200 in the local MkDocs preview.
- Browser checks at desktop (1440px) and mobile (390px) widths found no page-level horizontal overflow or broken article anchors.
- All 35 embedded images loaded successfully.

## Before Marking Ready

- [ ] Review the article and its figures.
- [ ] Add the article to the blog index and navigation.
- [ ] Add descriptive image alternative text.

## Summary by CodeRabbit

- **Documentation**
  - Added a new blog post covering RL-Kernel v0.1.0 and bitwise consistency between training and rollout.
  - Updated the blog index and site navigation to make the article easier to find.

- **Improvements**
  - Enhanced documentation pages with reliable inline and display mathematical notation rendering, including support during in-page navigation.

PR #427, opened 2026-09-18
fix(docs): fetch full history for strict documentation builds

maxiaosong1124:
## Summary
- Set `fetch-depth: 0` for the Pages build and the documentation job in CI.
- Keep strict MkDocs validation enabled and leave unrelated checkout steps unchanged.

## Root cause
After #418 merged, the Docs run failed because the Git revision date plugin warned about shallow checkout history. Strict mode treated that warning as fatal, so Pages deployment was skipped and the Blog index remained stale.

Failed deployment: https://github.com/RL-Align/RL-Kernel/actions/runs/35336980198

## Validation
- Reproduced the failure with a clean depth-1 clone and `GITHUB_ACTIONS=true`.
- Fetched full history in the same clone; `mkdocs build --strict -f mkdocs.yaml` then passed.
- Verified both documentation jobs use full history and retain strict mode.

Merging this PR will trigger a fresh Docs deployment from main.

## Summary by CodeRabbit

- **Chores**
  - Updated documentation and continuous integration workflows to fetch the repository’s full Git history during builds.


KJLdefeated (approved):
LGTM

Flink-ddd (approved):
LGTM

